PR #6, opened 2017-04-29
Allow downstream code to choose between running from RAM or Flash

whitequark:
This commit parameterizes the linker script to use the memory region
TARGET. This region can be defined in a downstream memory.x as follows:

    MEMORY { ... }
    REGION_ALIAS("TARGET", FLASH); /* or RAM */

Apart from RAM and FLASH, other memory regions could be used, e.g. in order to produce a bootloader. Because the entire process is orchestrated by the downstream crate's build.rs, it can be adjusted programmatically, without rebuilding any of the dependencies.

### DIFF
--- a/link.x
+++ b/link.x
@@ -2,7 +2,7 @@ INCLUDE memory.x
 
 SECTIONS
 {
-  .text ORIGIN(FLASH) :
+  .text ORIGIN(TARGET) :
   {
     /* Vector table */
     _VECTOR_TABLE = .;
@@ -17,7 +17,7 @@ SECTIONS
 
     *(.text.*);
     *(.rodata.*);
-  } > FLASH
+  } > TARGET
 
   .bss : ALIGN(4)
   {
@@ -31,7 +31,7 @@ SECTIONS
     _sdata = .;
     *(.data.*);
     _edata = ALIGN(4);
-  } > RAM AT > FLASH
+  } > RAM AT > TARGET
 
   _sidata = LOADADDR(.data);
 
@@ -44,14 +44,14 @@ SECTIONS
 }
 
 /* Do not exceed this mark in the error messages below                | */
-ASSERT(__exceptions - ORIGIN(FLASH) > 8, "
+ASSERT(__exceptions - ORIGIN(TARGET) > 8, "
 You must specify the exception handlers.
 Create a non `pub` static variable with type
 `cortex_m::exception::Handlers` and place it in the
 '.rodata.exceptions' section. (cf. #[link_section]). Apply the
 `#[used]` attribute to the variable to make it reach the linker.");
 
-ASSERT(__exceptions - ORIGIN(FLASH) == 0x40, "
+ASSERT(__exceptions - ORIGIN(TARGET) == 0x40, "
 Invalid '.rodata.exceptions' section.
 Make sure to place a static with type `cortex_m::exception::Handlers`
 in that section (cf. #[link_section]) ONLY ONCE.");


whitequark:
It is unfortunately not possible to undefine or redefine a region alias, so tutorials, etc, will have to be changed.

japaric:
Thanks for the PR, @whitequark.

> in order to produce a bootloader.

I'd love to support such setups with cortex-m-rt. But I'd like to see sketched
out how a bootloader would hook to this before landing this PR. I'm not familiar
with bootloader internals but the simplest way I can think of how it would work
is this:

- Bootloader boots from Flash in the normal way.
- set VTOR to 0x2000_0000
- call the entry point of the program loaded in RAM. So something like
  `(0x2000_0004 as fn() -> !)()`

But then I don't see how the program loaded in RAM would respect the 'initial
value for the stack pointer' stored in 0x2000_0000.

Then there's the question of whether this is flexible enough to support existing
bootloaders like the stm32 one.

And if there are only two options for the `REGION_ALIAS` argument (FLASH and
RAM) there perhaps this could be just a Cargo feature to switch between those
two.

> It is unfortunately not possible to undefine or redefine a region alias, so tutorials, etc, will have to be changed.

This seems fine since we have semver and documentation. Blog posts should pin to specific versions for reproducibility in any case.

whitequark:
> I'd love to support such setups with cortex-m-rt. But I'd like to see sketched
> out how a bootloader would hook to this before landing this PR.

I want to mention that bootloaders are not the main motivation for this PR. Mostly, I did it for debugging (I wasn't sure if it was openocd that's broken or flashing), and because from prior experience with stlink v1, flashing could take a long time, whereas loading to RAM was always extremely quick (this is not true on TM4C, but I imagine some STM32 users would be happier).

Anyway...

> But then I don't see how the program loaded in RAM would respect the 'initial
> value for the stack pointer' stored in 0x2000_0000.

... the bootloader could simply use a two-instruction trampoline that sets SP and PC. We could provide it as a part of cortex_m::asm, for that matter.

whitequark:
> And if there are only two options for the REGION_ALIAS argument (FLASH and
> RAM) there perhaps this could be just a Cargo feature to switch between those
> two.

I have initially thought about such design, and eventually opted against it. Here are my reasons:
  * The build process becomes quite a bit more complicated. The cortex-m-rt linker script is hardcoded in the `.cargo/config` of the downstream crate. Changing to change the build configuration is highly undesirable as it is effortful to do right and it causes a rebuild of every dependency. So, we have to, instead of copying `link.x` as it is, adjust it in the build script. This makes cortex-m-rt harder to understand.
  * The features are not a particularly good mechanism to express this. If we have two features, `ram` and `flash`, then half of the possible combinations is meaningless.
  * Most importantly, there's no reason there should be *just* two memory areas. The uC you've mentioned in the quickstart post has two RAM areas; I know there are uCs that have discontiguous flash; even on uCs with contiguous flash, it is a perfectly valid desire to split it manually, such as to add a bootloader or selectable firmware.

My REGION_ALIAS method has none of the issues above and requires only minimal changes to cortex-m-rt.

whitequark:
> Then there's the question of whether this is flexible enough to support existing
> bootloaders like the stm32 one.

I'm not sure what you are referring to.

japaric:
> ... the bootloader could simply use a two-instruction trampoline that sets SP and PC.

With just that changes to the vector table in the source wouldn't carry over, right? The uC would still be using the vector table at 0x0 / Flash.

> The cortex-m-rt linker script is hardcoded in the .cargo/config of the downstream crate.

This is not entirely correct. .cargo/config must always be in the top crate / Cargo configuration. (Cargo ignores the .cargo/config of dependencies). The linker script name is hardcoded to link.x if you are using cortex-m-rt so there's no need to ever change the linker script settings in .cargo/config.

> Changing to change the build configuration is highly undesirable as it is effortful to do right and it causes a rebuild of every dependency.

I don't see how it would cause a full rebuild. cortex-m-rt can always be used as a direct dependency of the application crate and that's how I think it should be used (but I can't foresee all the possible use cases). In that case, changing a Cargo feature in cortex-m-rt only causes that crate to be rebuilt, not the entire dependency graph.

> The features are not a particularly good mechanism to express this.

I was thinking of just a single feature then we'd have two possibilities.

> Most importantly, there's no reason there should be just two memory areas

I know.

> I know there are uCs that have discontiguous flash; even on uCs with contiguous flash, it is a perfectly valid desire to split it manually, such as to add a bootloader or selectable firmware.

Which is why it'd like to see such use cases sketched out to make sure the changes proposed here support those use cases.

So, in my mind, we either make the linker script flexible enough to support elaborated memory layouts or we just support a few simpler memory layouts (what we have right now for deploys, and everything on RAM for development) and tell users how to plug their own linker script for more elaborated setups ("disable the 'linker-script' feature in cortex-m-rt, use this build script that copies link.x over, start from this link.x file)

whitequark:
> With just that changes to the vector table in the source wouldn't carry over, right? The uC would still be using the vector table at 0x0 / Flash.

Sure, you've mentioned updating VTOR already which is why I did not repeat it.

> This is not entirely correct. .cargo/config must always be in the top crate / Cargo configuration.

Yeah, that's what I meant by "downstream crate", bad phrasing.

Anyway. I would still like to have this ability, but I cannot justify the effort to make this PR much more generic, and I do not like the Cargo feature suggestion enough to work on it, so I am going to close it for now.

PR #7, opened 2017-04-30
Mark .debug_gdb_scripts as non-allocatable

whitequark:
See the discussion in https://github.com/rust-lang/rust/pull/41627
for an explanation of why this is necessary.

### DIFF
--- a/link.x
+++ b/link.x
@@ -35,6 +35,17 @@ SECTIONS
 
   _sidata = LOADADDR(.data);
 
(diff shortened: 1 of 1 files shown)


japaric:
Obscure stuff... Thanks for looking into this @whitequark!

@homunkulus r+

whitequark:
@japaric looks like @homunkulus is broken?

japaric:
Er, homu is not set up for this repo because there's no CI. Did the r+ out of habit without checking.

PR #8, opened 2017-05-25
reorder the linker sections; make reset_handler the entry point

japaric:
Before this commit the .rodata sections of the inputs were placed in the output
.text section. As the vector table is in the .rodata of the cortex-m-rt crate
this placed the vector table first in the output .text section:

``` console
$ arm-none-eabi-objdump -Cd hello
Disassembly of section .text:

08000000 <_VECTOR_TABLE>:
 8000000:       10002000        .word   0x10002000

08000004 <cortex_m_rt::RESET_HANDLER>:
 8000004:       08000405                                ....
(..)

$ arm-none-eabi-readelf -h hello | grep Entry
  Entry point address:               0x8000000
```

With this memory layout using `load` and then `step` on a GDB session made the
processor execute the vector table (PC = 0x08000000).

This commit places the .rodata sections of the inputs into the output .rodata
section, and also moves the reset_handler code to the start the output .text
section. This turns the reset_handler into the entry point of the program.

``` console
$ arm-none-eabi-objdump -Cd hello
Disassembly of section .text:

08000410 <cortex_m_rt::reset_handler>:
(..)

$ arm-none-eabi-readelf -h hello | grep Entry
  Entry point address:               0x8000410
```

With these changes the GDB commands `load` and `step` work correctly: `load`
sets the PC = reset_handler.

Also the `size` output is now more helpful:

``` console
$ arm-none-eabi-size -Ax hello
section                 size         addr
.vector_table          0x400    0x8000000
.text                  0x120    0x8000400
.rodata                  0xd    0x8000520
.bss                     0x4   0x20000000
.data                    0x4   0x20000004
```

---

r? @whitequark

With this we should be able to drop the `monitor reset halt` command from the
`.gdbinit` file in cortex-m-quickstart.

### DIFF
--- a/link.x
+++ b/link.x
@@ -2,21 +2,32 @@ INCLUDE memory.x
 
 SECTIONS
 {
-  .text ORIGIN(FLASH) :
+  .vector_table ORIGIN(FLASH) :
   {
     /* Vector table */
-    _VECTOR_TABLE = .;
+    _svector_table = .;
     LONG(_stack_start);
 
-    KEEP(*(.rodata.reset_handler));
+    KEEP(*(.vector_table.reset_handler));
 
-    KEEP(*(.rodata.exceptions));
+    KEEP(*(.vector_table.exceptions));
     _eexceptions = .;
 
-    KEEP(*(.rodata.interrupts));
+    KEEP(*(.vector_table.interrupts));
     _einterrupts = .;
+  } > FLASH
+
+  .text : ALIGN(4)
+  {
+    /* Put reset handler first in .text section so it ends up as the entry */
+    /* point of the program. */
+    KEEP(*(.text.reset_handler));
 
     *(.text .text.*);
+  } > FLASH
+
+  .rodata : ALIGN(4)
+  {
     *(.rodata .rodata.*);
   } > FLASH
 
--- a/link.x
+++ b/link.x
@@ -21,7 +21,7 @@ SECTIONS
   {
     /* Put reset handler first in .text section so it ends up as the entry */
     /* point of the program. */
-    KEEP(*(.text.reset_handler));
+    KEEP(*(.reset_handler));
 
     *(.text .text.*);
   } > FLASH
--- a/link.x
+++ b/link.x
@@ -10,10 +10,10 @@ SECTIONS
 
     KEEP(*(.vector_table.reset_handler));
 
-    KEEP(*(.vector_table.exceptions));
+    KEEP(*(.rodata.exceptions));
     _eexceptions = .;
 
-    KEEP(*(.vector_table.interrupts));
+    KEEP(*(.rodata.interrupts));
     _einterrupts = .;
   } > FLASH
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -213,7 +213,7 @@ static RESET_HANDLER: unsafe extern "C" fn() -> ! = reset_handler;
 
 #[allow(dead_code)]
 #[cfg(feature = "exceptions")]
-#[link_section = ".vector_table.exceptions"]
+#[link_section = ".rodata.exceptions"]
 #[used]
 static EXCEPTIONS: exception::Handlers = exception::Handlers {
     ..exception::DEFAULT_HANDLERS
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -189,7 +189,7 @@ extern "C" {
 /// The reset handler
 ///
 /// This is the entry point of all programs
-#[link_section = ".text.reset_handler"]
+#[link_section = ".reset_handler"]
 unsafe extern "C" fn reset_handler() -> ! {
     ::r0::zero_bss(&mut _sbss, &mut _ebss);
     ::r0::init_data(&mut _sdata, &mut _edata, &_sidata);
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -189,6 +189,7 @@ extern "C" {
 /// The reset handler
 ///
 /// This is the entry point of all programs
+#[link_section = ".text.reset_handler"]
 unsafe extern "C" fn reset_handler() -> ! {
     ::r0::zero_bss(&mut _sbss, &mut _ebss);
     ::r0::init_data(&mut _sdata, &mut _edata, &_sidata);
@@ -207,12 +208,12 @@ unsafe extern "C" fn reset_handler() -> ! {
 
 #[allow(dead_code)]
 #[used]
-#[link_section = ".rodata.reset_handler"]
+#[link_section = ".vector_table.reset_handler"]
 static RESET_HANDLER: unsafe extern "C" fn() -> ! = reset_handler;
 
 #[allow(dead_code)]
 #[cfg(feature = "exceptions")]
-#[link_section = ".rodata.exceptions"]
+#[link_section = ".vector_table.exceptions"]
 #[used]
 static EXCEPTIONS: exception::Handlers = exception::Handlers {
     ..exception::DEFAULT_HANDLERS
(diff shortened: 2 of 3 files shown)


whitequark:
The underlying idea of this change looks good to me; I've hit this issue sometime ago (while investigating uploading code into SRAM) but never went to the bottom of it.

I'm far less certain about the section names. The name convention of `.text.something` or `.rodata.something` originates from the hosted ELF targets, and its purpose is a weakly structured form of communication between the various compilers and the system linker. E.g. in my system (try `ld --verbose`):

```
  .text           :
  {
    *(.text.unlikely .text.*_unlikely .text.unlikely.*)
    *(.text.exit .text.exit.*)
    *(.text.startup .text.startup.*)
    *(.text.hot .text.hot.*)
    *(.text .stub .text.* .gnu.linkonce.t.*)
    /* .gnu.warning sections are handled specially by elf32.em.  */
    *(.gnu.warning)
  }
```
Essentially, so long as there is a `*(.text.*)` somewhere, the compilers can introduce new conventions without hard breakage. This also incidentally allows `-ffunction-sections` to work (I think this convention predates the flag).

In case of cortex-m-rt we lack this compatibility hazard, and so I think there is no good reason to put something into a `.text.something` or `.rodata.something` section when we then explicitly specify that in the linker script. I would prefer to have an explicit `.vectors` section in the output, such that `arm-none-eabi-size` would be more useful on the final ELF product; the vector table shouldn't count against the data size, for one. This is what basically every other toolchain does with their linker scripts.

There are also a few minor issues for which I'll leave inline comments.

whitequark:
Couldn't leave inline comments so #9 and #10.

japaric:
> such that arm-none-eabi-size would be more useful on the final ELF product;

I should note that cortex-m-rt in its current form, this PR, and the modified version of this PR with an output .vectors section all produce the same arm-none-eabi-size output. That is the vector table is counted as "text", not as "data" or "bss".

>  I think there is no good reason to put something into a .text.something or .rodata.something section

If I understand you correctly we should *not* use, for example, `#[linker_section = ".rodata.exceptions"]` in Rust code because the toolchain / ELF standard may use use the ".rodata.exceptions" section in the future for some feature, or assign some other meaning to it. Instead we should use something like `#[linker_section = ".vectors.exceptions"]`. And you comment about an *output* `.vectors* section was mainly due to the concern of arm-none-eabi-size not behaving correctly? The section `.vectors.exceptions` wouldn't make it to the output binary because the linker script will instruct the linker to produce a single `.vectors` section in the output binary.

whitequark:
> I should note that cortex-m-rt in its current form, this PR, and the modified version of this PR with an output .vectors section all produce the same arm-none-eabi-size output.

Try the `-A` option.

> If I understand you correctly we should not use, for example, #[linker_section = ".rodata.exceptions"] in Rust code because the toolchain / ELF standard may use use the ".rodata.exceptions" section in the future for some feature, or assign some other meaning to it.

No, I meant there is no argument *for* using this name for the section. We control the entire toolchain (the linker doesn't make judgement of section names other than via the builtin script) so there's no argument *against* based on what the ELF standard or toolchain does.

> Instead we should use something like #[linker_section = ".vectors.exceptions"].

I think having separate sections for vectors and not squishing them into `.rodata` is just handy for anyone working with the files on that level, which I had to do on quite a few occasions. The `arm-none-eabi-size` improvement is perhaps the most visible change.

> The section .vectors.exceptions wouldn't make it to the output binary because the linker script will instruct the linker to produce a single .vectors section in the output binary.

Yeah sure, changing *that* name is just for consistency with the output product.

pftbest:
What is wrong with doing it this way, like any other linker script from C world? Is there a reason why we need to put all sections in one place (eg. in .text)?
```c
SECTIONS
{
  .vectors ORIGIN(FLASH):
  {
    /* Vector table */
    _VECTOR_TABLE = .;
    LONG(_stack_start);

    KEEP(*(.vectors.reset_handler));
    KEEP(*(.vectors.exceptions));
    __exceptions = .;

    KEEP(*(.vectors.interrupts));
    __interrupts = .;
  } > FLASH

  .text : ALIGN(4)
  {
    /* Put reset handler first in .text section so it ends up as the entry */
    /* point of the program. */
    KEEP(*(.text.reset_handler));
    *(.text)
    *(.text*)
  } > FLASH

  .rodata : ALIGN(4)
  {
    *(.rodata)
    *(.rodata*)
  } > FLASH

  .data : ALIGN(4)
  {
    _sdata = .;
    *(.data .data.*);
    _edata = ALIGN(4);
  } > RAM AT > FLASH

  .bss : ALIGN(4)
  {
    _sbss = .;
    *(.bss .bss.*);
    _ebss = ALIGN(4);
  } > RAM

```

whitequark:
@pftbest this is exactly what I'm suggesting fwiw

japaric:
> Try the -A option.

TIL. I agree that's more helpful with the .vector split.

I have made the requested changes but I have picked .vector_table instead of .vectors. Does that rename seems good to you?

link.x
```diff
@@ -2,21 +2,32 @@ INCLUDE memory.x
 
 SECTIONS
 {
-  .text ORIGIN(FLASH) :
+  .vector_table ORIGIN(FLASH) :
   {
     /* Vector table */
-    _VECTOR_TABLE = .;
+    _svector_table = .;
```
japaric: this is not used in the linker script at all. It's just to make the `objdump` output slightly nicer: you get `<_svector_table>` instead of `<cortex_m_rt::RESET_HANDLER-0x4>`.

src/lib.rs
```diff
@@ -189,6 +189,7 @@ extern "C" {
 /// The reset handler
 ///
 /// This is the entry point of all programs
+#[link_section = ".text.reset_handler"]
```
japaric: Q: should we rename this linker section to `.reset_handler`? The name is an implementation detail of the cortex-m-rt as no other crate is supposed to use this section. Now I wonder if `.text.reset_handler` makes more likely that the linker section may get misused. For example some C code could place code in this section if a function is named `reset_handler` and `-ffunction-sections` is used, or something along those lines.

japaric:
Also check my inline comments; I left a question there about .text.reset_handler.

pftbest:
looks good to me, i don't mind the rename. Also i agree that having `.reset_handler` instead of `.text.reset_handler` is less likely to break.

src/lib.rs
whitequark: A: yes

link.x
whitequark: Shouldn't that be something like `_pfstack` or something?

whitequark:
@japaric I prefer `.vectors` (after all we don't call `.bss` `.uninit_data`) but it's not a hill I will die on

link.x
pftbest: At least to me it's not immediately obvious what `_pfstack` means.

japaric: I give up. What does `_pfstack` stand for? 🙂 

I chose _svector_table because of the parallel to _sbss and _sdata which are named after their output section names

japaric:
> I prefer .vectors

I chose .vector_table because that's how the ARM documentation calls it. Also not everything in .vector_table is a vector; the first thing on the vector table is the initial SP value.

link.x
whitequark: well the only thing actually offset from the `_svector_table` symbol is the initial SP, right? So, pointer to `_fstack`, so pointer to start of stack. I admit this is a pretty bad initialism if none of you got it :) I have nothing, sure, `_svectors` sounds good enough

whitequark:
> I chose .vector_table because that's how the ARM documentation calls it. Also not everything in .vector_table is a vector; the first thing on the vector table is the initial SP value.

Ack.

japaric:
Thanks for input @whitequark and @pftbest. I'll postpone the rename of the input sections .rodata.{interrupts,exceptions} to .vector_table.* as it would break crates like cortex-m-quickstart and cortex-m-rtfm. That would let us publish a new patch version with these changes today.

PR #9, opened 2017-05-26
Explicitly place non-suffixed sections

whitequark:
The `*(.data.*);` command does not place the `.data` section, which may lead
to incorrect placement of symbols, etc. I have not verified that the current
linker script leads to broken products, but if not it is by accident.

### DIFF
--- a/link.x
+++ b/link.x
@@ -15,21 +15,21 @@ SECTIONS
     KEEP(*(.rodata.interrupts));
     __interrupts = .;
 
-    *(.text.*);
-    *(.rodata.*);
+    *(.text .text.*);
+    *(.rodata .rodata.*);
   } > FLASH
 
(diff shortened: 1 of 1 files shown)


japaric:
Thanks @whitequark. The current linker script *may* produce broken binaries if linking against C code compiled without -fdata-sections / -ffunction-sections. I haven't seen that thing happen in practice since I don't usually link to C libraries.

whitequark:
Does rustc have a hard guarantee that it will not produce a .text section? Even with LTO enabled?

japaric:
I recall seeing some code in the rust backend that makes LLVM do the equivalent of -fdata-sections and -ffunction-sections. 

> Does rustc have a hard guarantee that it will not produce a .text section? Even with LTO enabled?

That I do not know.

PR #10, opened 2017-05-26
Rename __exceptions/__interrupts symbols to reduce confusion

whitequark:
Right now these symbols get placed *after* their respective contents,
but not named to indicate that.

### DIFF
--- a/link.x
+++ b/link.x
@@ -9,11 +9,12 @@ SECTIONS
     LONG(_stack_start);
 
     KEEP(*(.rodata.reset_handler));
+
     KEEP(*(.rodata.exceptions));
-    __exceptions = .;
+    _eexceptions = .;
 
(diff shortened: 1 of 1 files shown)


whitequark:
I've also renamed them to `_sexceptions` and `_sinterrupts`, following the convention for `_sbss` etc.

There's also the question of usefulness; do we even need them? Doesn't seem like there is much harm to have them, and maybe a little useful when staring at objdump, I guess. Conversely, should we maybe add `_stext`/`_etext` for consistency?

japaric:
> There's also the question of usefulness; do we even need them?

Their main use at the assertions at the bottom of the linker script. Which would need to be updated to accomodate the rename in this PR.

japaric:
Thanks!

PR #14, opened 2017-05-29
[WIP] Optional .reserved region between .vector_table and .text

### LINKED ISSUE #13: Reserve flash memory after .vector_table
There should be some way to reserve a section of flash so nothing will be placed there. Particularly right after .vector_table

Some use cases for this feature
- Some MCU stores configurations right after .vector_table (for instance the NXP S32K series store flash configuration from 0x400-0x40C)
- Reserving space for an alternative program (like a bootloader) 

I guess the simple solution will be to introduce a text_offset (or text_start) symbol which can be defined in the memory.x and if defined be used to place a reserved region between .vector_table and .text (or simply offsetting .text)

Alternatively are there any semi-simple ways to implement support for reserving a region of flash(/ram) in an arbitrary location without breaking the way the memory.x files work?

kjetilkjeka:
A simple way to fix #13 

Set text_offset in memory.x to override default value of 0

Since text_offset doesn't need to be set this is backwards compatible.

Is a simple fix like this wanted or is there a better solution that is preffered.

Is the form text_offset or text_start (or somthing similar) the preffered way to set this offset size?

### DIFF
--- a/link.x
+++ b/link.x
@@ -18,14 +18,8 @@ SECTIONS
   } > FLASH
 
   PROVIDE(_stext = _einterrupts);
-  _reserved_size = _stext - _einterrupts;
-
-  .reserved : ALIGN(4)
-  {
-    . += _reserved_size;
-  } > FLASH
   
-  .text : ALIGN(4)
+  .text _stext : ALIGN(4)
   {
     /* Put reset handler first in .text section so it ends up as the entry */
     /* point of the program. */
--- a/link.x
+++ b/link.x
@@ -1,3 +1,5 @@
+text_offset = 0;
+
 INCLUDE memory.x
 
 SECTIONS
@@ -15,6 +17,11 @@ SECTIONS
 
     KEEP(*(.rodata.interrupts));
     _einterrupts = .;
+  } > FLASH
+
+  .reserved : ALIGN(4)
+  {
+    . += text_offset;
   } > FLASH
 
   .text : ALIGN(4)
--- a/link.x
+++ b/link.x
@@ -17,8 +17,8 @@ SECTIONS
     _einterrupts = .;
   } > FLASH
 
-  PROVIDE(_text_start = _einterrupts);
-  _reserved_size = _text_start - _einterrupts;
+  PROVIDE(_stext = _einterrupts);
+  _reserved_size = _stext - _einterrupts;
 
   .reserved : ALIGN(4)
   {
@@ -100,10 +100,10 @@ ASSERT(_einterrupts - _eexceptions <= 0x3c0, "
 There can't be more than 240 interrupt handlers.
 Fix the '.rodata.interrupts' section. (cf. #[link_section])");
 
-ASSERT(_einterrupts <= _text_start, "
+ASSERT(_einterrupts <= _stext, "
 The '.text' section can't be placed inside '.vector_table' section.
-Set '_text_start' to an adress greater than '_einterrupts'");
+Set '_stext' to an adress greater than '_einterrupts'");
 
-ASSERT(_text_start < ORIGIN(FLASH) + LENGTH(FLASH), "
+ASSERT(_stext < ORIGIN(FLASH) + LENGTH(FLASH), "
 The '.text' section must be placed inside the FLASH memory
-Set '_text_start' to an adress smaller than 'ORIGIN(FLASH) + LENGTH(FLASH)");
+Set '_stext' to an adress smaller than 'ORIGIN(FLASH) + LENGTH(FLASH)");
--- a/link.x
+++ b/link.x
@@ -106,4 +106,4 @@ Set '_stext' to an adress greater than '_einterrupts'");
 
 ASSERT(_stext < ORIGIN(FLASH) + LENGTH(FLASH), "
 The '.text' section must be placed inside the FLASH memory
-Set '_stext' to an adress smaller than 'ORIGIN(FLASH) + LENGTH(FLASH)");
+Set '_stext' to an address smaller than 'ORIGIN(FLASH) + LENGTH(FLASH)");
--- a/link.x
+++ b/link.x
@@ -99,3 +99,11 @@ Create a non `pub` static variable and place it in the
 ASSERT(_einterrupts - _eexceptions <= 0x3c0, "
 There can't be more than 240 interrupt handlers.
 Fix the '.rodata.interrupts' section. (cf. #[link_section])");
+
+ASSERT(_einterrupts <= _text_start, "
+The '.text' section can't be placed inside '.vector_table' section.
+Set '_text_start' to an adress greater than '_einterrupts'");
+
+ASSERT(_text_start < ORIGIN(FLASH) + LENGTH(FLASH), "
+The '.text' section must be placed inside the FLASH memory
+Set '_text_start' to an adress smaller than 'ORIGIN(FLASH) + LENGTH(FLASH)");
--- a/link.x
+++ b/link.x
@@ -1,5 +1,3 @@
-text_offset = 0;
-
 INCLUDE memory.x
 
 SECTIONS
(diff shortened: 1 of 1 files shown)


japaric:
I think a text_start approach may be less error prone. You can't be sure that the .vector_table section will be big enough to match the real Flash memory region where the vector table is supposed to be located (specially if the user is not using a svd2rust generated device crate and they are creating the .rodata.interrupts section manually).

OTOH, with the text_start approach it is possible to pick a value that will make the .text section and the .vector_table section overlap *but* we can add an assertion to make sure that doesn't link: `assert(_einterrupts <= text_start); assert(text_start < ORIGIN(FLASH) + LENGTH(FLASH));`.

kjetilkjeka:
The way it works now is that you can choose to set _text_start from the memory.x file and if set (and a sensible value) a .reserved section will be placed as padding between .vector_table and .text

I've added the assertions but since they're placed in the bottom it will fail before reaching them.

If _text_start is inside .vector_table the error message given will be: "link.x:25 cannot move location counter backwards (from 0000000000000400 to 0000000000000100)"

If _text_start is outside flash: "section `.reserved' will not fit in region `FLASH'"

Would it be prefered to force the assertions to give the specialized error messages by moving them within the section definitions? Or is this fine?

link.x
```diff
@@ -17,6 +17,14 @@ SECTIONS
     _einterrupts = .;
   } > FLASH
 
+  PROVIDE(_text_start = _einterrupts);
```
japaric: nit: could you rename _text_start to _start? That's the convention we have for symbols that point to section boundaries.

link.x
```diff
@@ -91,3 +99,11 @@ Create a non `pub` static variable and place it in the
...
 There can't be more than 240 interrupt handlers.
 Fix the '.rodata.interrupts' section. (cf. #[link_section])");
+
+ASSERT(_einterrupts <= _text_start, "
+The '.text' section can't be placed inside '.vector_table' section.
+Set '_text_start' to an adress greater than '_einterrupts'");
+
+ASSERT(_text_start < ORIGIN(FLASH) + LENGTH(FLASH), "
+The '.text' section must be placed inside the FLASH memory
+Set '_text_start' to an adress smaller than 'ORIGIN(FLASH) + LENGTH(FLASH)");
```
japaric: typo: "adress"

link.x
japaric: > _start

err, I meant _stext

link.x
```diff
@@ -17,6 +17,14 @@ SECTIONS
     _einterrupts = .;
   } > FLASH
 
+  PROVIDE(_stext = _einterrupts);
+  _reserved_size = _stext - _einterrupts;
+
+  .reserved : ALIGN(4)
+  {
+    . += _reserved_size;
```
japaric: nit: could you change the RHS to `(_stext - _einterrupts)` and remove the `_reserved_size` symbol / variable from this file? One can bind to all the symbols / variables used in this file so it's better to keep their numbers small to reduce the chance of misuse.

``` rust
extern {
    static _reserved_size: u32;
}
let address = unsafe { &_reserved_size as *const _ as usize };
// ...
```

link.x
```diff
@@ -91,3 +99,11 @@ Create a non `pub` static variable and place it in the
 ASSERT(_einterrupts - _eexceptions <= 0x3c0, "
 There can't be more than 240 interrupt handlers.
 Fix the '.rodata.interrupts' section. (cf. #[link_section])");
+
+ASSERT(_einterrupts <= _stext, "
+The '.text' section can't be placed inside '.vector_table' section.
+Set '_stext' to an adress greater than '_einterrupts'");
```
japaric: Too bad we can't format _einterrupts here. An address like `0x0800_0200` would be more useful than  just _einterrupts.

kjetilkjeka:
@japaric considering the error messages. When I test on with _stext = 0x20d (less than .vector_table the full linker error i run into is:

    error: linking with `arm-none-eabi-ld` failed: exit code: 1
      |
      = note: "arm-none-eabi-ld" "-L" "/home/kjeka/.xargo/lib/rustlib/thumbv7em-none-eabihf/lib" 
    "/home/kjeka/test/s32k144evb/target/thumbv7em-none-eabihf/debug/examples/light-
    ec6131f661446433.0.o" "-o" "/home/kjeka/test/s32k144evb/target/thumbv7em-none-
    eabihf/debug/examples/light-ec6131f661446433" "--gc-sections" "-L" 
    "/home/kjeka/test/s32k144evb/target/thumbv7em-none-eabihf/debug/deps" "-L" 
    "/home/kjeka/test/s32k144evb/target/debug/deps" "-L" 
    "/home/kjeka/test/s32k144evb/target/thumbv7em-none-eabihf/debug/build/s32k144evb-
    9cc591d239f39664/out" "-L" "/home/kjeka/test/s32k144evb/target/thumbv7em-none-
    eabihf/debug/build/cortex-m-rt-323f7f33db0a273a/out" "-L" "/home/kjeka/.xargo/lib/rustlib/thumbv7em-
    none-eabihf/lib" "-Bstatic" "/tmp/rustc.Ah161pWyuOUA/libcompiler_builtins-48645c160d69be81.rlib" "-
    Tlink.x" "-Bdynamic"
      = note: link.x:25 cannot move location counter backwards (from 0000000000000400 to 
    000000000000020d)


But when testing with a value outside flash (_stext = 0xffff0000) it seems like the assertions are evaluated.

    error: linking with `arm-none-eabi-ld` failed: exit code: 1
      |
      = note: "arm-none-eabi-ld" "-L" "/home/kjeka/.xargo/lib/rustlib/thumbv7em-none-eabihf/lib"             
    "/home/kjeka/test/s32k144evb/target/thumbv7em-none-eabihf/debug/examples/light-    
    ec6131f661446433.0.o" "-o" "/home/kjeka/test/s32k144evb/target/thumbv7em-none-
    eabihf/debug/examples/light-ec6131f661446433" "--gc-sections" "-L" 
    "/home/kjeka/test/s32k144evb/target/thumbv7em-none-eabihf/debug/deps" "-L" 
    "/home/kjeka/test/s32k144evb/target/debug/deps" "-L" 
    "/home/kjeka/test/s32k144evb/target/thumbv7em-none-eabihf/debug/build/s32k144evb-
    9cc591d239f39664/out" "-L" "/home/kjeka/test/s32k144evb/target/thumbv7em-none-
    eabihf/debug/build/cortex-m-rt-323f7f33db0a273a/out" "-L" "/home/kjeka/.xargo/lib/rustlib/thumbv7em-
    none-eabihf/lib" "-Bstatic" "/tmp/rustc.G618pSJeUERd/libcompiler_builtins-48645c160d69be81.rlib" "-
    Tlink.x" "-Bdynamic"
      = note: arm-none-eabi-ld: /home/kjeka/test/s32k144evb/target/thumbv7em-none-
    eabihf/debug/examples/light-ec6131f661446433 section `.reserved' will not fit in region `FLASH'
              arm-none-eabi-ld: 
              The '.text' section must be placed inside the FLASH memory
              Set '_stext' to an address smaller than 'ORIGIN(FLASH) + LENGTH(FLASH)
              arm-none-eabi-ld: region `FLASH' overflowed by 4294381036 bytes
    

For some reason the assertions won't be evaluated after the location pointer has been decremented. To confirm this I also tested with: ```ASSERT(1>2, "print this string");``` and the string is not printed for _stext = 0x20d.

japaric:
Does moving the assertion before the .reserved section help? Does the assertion message show in that case?

kjetilkjeka:
By ditching the .reserved section and setting start of .text manually i was able to fix it.

I guess the .reserved section isn't needed since we can be sure that nothing is placed between .vector_table and .text?

kjetilkjeka:
Moving the assertion won't work since the linker doesn't allow me to have an assertion inside the SECTIONS and _einterrupts isn't defined before the start of SECTIONS.

japaric:
Thanks @kjetilkjeka! This looks great now.

PR #19, opened 2017-06-25
[RFC] default panic! to abort, drop panic-* features, add panic_fmt! macro

japaric:
# What

I propose we drop the panic-* features in this repo.

# Why

If we want to use this crate *not* as a direct dependency of the top crate then,
IMO, it's better to have less features otherwise whatever other crate brings
this crate into the dependency graph will have to re-export a bunch of Cargo
features. The re-exports aggravate if there are more crates between the top
crate and the rt crate.

# How

We'll drop the panic-* features and change the panic_fmt lang item to be just a
call to `intrinsics::abort`. We'll also add a `panic_fmt!` macro to make it
easier and type safer to override the `panic_fmt` implementation.

# Downside: regression in debugging experience

Today is easy to turn panic!s into formatted strings: you can just enable the
"panic-over-semihosting" feature -- easy, assuming that cortex-m-rt is a direct
dependency of the top crate. With this proposed change you'll have to define the
formatting implementation yourself.

Do note that you can, at least, recover the filename and line information using
a debugger with the proposed default `panic!` equals abort implementation. The
abort will cause a hard fault exception and the default exception handler gives
you access to the stacked register and the r0 register points to a tuple that
contains the filename and line information. With code:

``` rust
fn main() {
    panic!("foo");
}
```

``` console
> # GDB

> continue
cortex_m_rt::default_handler (_sr=0x20004fd0) at (..)/cortex-m-rt/src/lib.rs:263

> p/x _sr.r0
$1 = 0x8000204

> x/5wx 0x8000204
0x8000204:      0x080001f4      0x00000003     0x080001f7      0x0000000b
0x8000214:      0x0000000d

> x/3c 0x080001f4
0x80001f4 <str.0>:      102 'f' 111 'o' 111 'o'

> x/11c 0x080001f7
0x80001f7 <str.1>:      115 's' 114 'r' 99 'c'  47 '/'  109 'm' 97 'a'  105 'i' 110 'n'
0x80001ff <str.1+8>:    46 '.'  114 'r' 115 's'

> # line number is at 0x8000214
```

# Alternatives

We could add an opt-out `panic_fmt` Cargo feature to this crate. When that
feature is enabled this crate will provide a `panic_fmt` implementation, and
when not the user will be responsible for defining the `panic_fmt` lang item
through some other way.

# Unresolved questions

- Take a closure or a path to function in the `panic_fmt!` macro?

Thoughts? @whitequark @pftbest

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ keywords = ["arm", "cortex-m", "runtime", "startup"]
 license = "MIT OR Apache-2.0"
 name = "cortex-m-rt"
 repository = "https://github.com/japaric/cortex-m-rt"
-version = "0.2.4"
+version = "0.3.0"
 
 [dependencies]
 r0 = "0.2.1"
@@ -22,3 +22,5 @@ default = ["exceptions", "linker-script"]
 exceptions = ["cortex-m"]
 # generic linker script
 linker-script = []
+# provides a panic_fmt implementation that calls the abort instruction (`udf 0xfe`)
+abort-on-panic = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,17 +16,9 @@ r0 = "0.2.1"
 optional = true
 version = "0.2.7"
 
-[dependencies.cortex-m-semihosting]
-optional = true
-version = "0.1.3"
-
 [features]
 default = ["exceptions", "linker-script"]
 # service all exceptions using the default handler
 exceptions = ["cortex-m"]
 # generic linker script
 linker-script = []
-# prints panic messages to the ITM
-panic-over-itm = ["cortex-m"]
-# prints panic messages to the host stdout
-panic-over-semihosting = ["cortex-m-semihosting"]
--- a/src/lang_items.rs
+++ b/src/lang_items.rs
@@ -6,32 +6,21 @@ unsafe extern "C" fn panic_fmt(
     _file: &'static str,
     _line: u32,
 ) -> ! {
-    match () {
-        #[cfg(feature = "panic-over-itm")]
-        () => {
-            use cortex_m::itm;
-            use cortex_m::peripheral::ITM;
+    ::core::intrinsics::abort()
+}
 
-            let port = &(*ITM.get()).stim[0];
-            iprint!(port, "panicked at '");
-            itm::write_fmt(port, _args);
-            iprintln!(port, "', {}:{}", _file, _line);
+#[macro_export]
+macro_rules! panic_fmt {
+    ($f:expr) => {
+        #[export_name = "rust_begin_unwind"]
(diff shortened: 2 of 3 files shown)


whitequark:
I'm not particularly happy about the reduction in ergonomics that results from this. This "regression in debugging experience" is not a downside, it will make the crate downright unusable unless the user puts in time to learn how lang items and such work, which is not reasonable to expect.

How exactly would an end user employ a particular panicking strategy?

> it's better to have less features otherwise whatever other crate brings
> this crate into the dependency graph will have to re-export a bunch of Cargo
> features

Why? The root crate can just activate the feature directly by depending on cortex-m-rt itself.

japaric:
> unless the user puts in time to learn how lang items and such work, which is
> not reasonable to expect.

Panicking is a very important component of the behavior of an embedded
application. I'd rather have people learn about customizing it upfront rather
than rely on a default that will likely end up doing the wrong thing when their
application fails (in production because Murphy's law).

Probably once `panic_fmt` lang item becomes stable (currently being worked on)
and well documented, I'll just remove the lang item from this crate and leave it
up to the application writer to supply the item. After all panicking behavior is
very application specific in no-std context.

> How exactly would an end user employ a particular panicking strategy?

My preference would be to have the end user directly supply the lang item
actually. Specially if the interface is stable.

> Why? The root crate can just activate the feature directly by depending on
> cortex-m-rt itself.

The more places the rt crate appears in the higher the chances of two
incompatible features being enabled: e.g. `panic-over-itm` and
`panic-over-semihosting` in the current state of this crate. This effectively
makes two crates that depend, directly or indirectly, on the rt crate non
composable.

---

I think I'm currently leaning towards *not* including the lang item by default,
but have an opt-in Cargo feature that provides some default (e.g. abort or maybe
something more debug oriented). The application writer can then add a direct
dependency to the rt crate to enable that default. The quickstart template can
include the rt crate with that feature enabled.

Furthermore, we should discourage libraries that depend on the rt crate from
enabling the feature that provides the lang item because that would lock all the
user of the libraries to a certain panicking behavior. The rt crate should *not*
provide several Cargo features to customize the panicking behavior as that opens
the door to the composability problem mentioned above.

whitequark:
> Probably once panic_fmt lang item becomes stable (currently being worked on)
> and well documented, I'll just remove the lang item from this crate and leave it
> up to the application writer to supply the item.

This is not a bad solution but I was under impression that lang items would *never* be stabilized.

japaric:
Only panic_fmt is being considered for stabilization since it's a blocker for doing no-std / embedded application development on stable.

whitequark:
OK. No objection then.

japaric:
I'm going to land this so I can test it with #18 but I'm not going to make a release until #18 lands so if anyone has an objection on this PR you can still comment here.

PR #20, opened 2017-06-29
enable the FPU before main if the target has an FPU

japaric:
cc @whitequark

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -189,20 +189,33 @@ unsafe extern "C" fn reset_handler() -> ! {
     r0::init_data(&mut _sdata, &mut _edata, &_sidata);
 
     match () {
+        #[cfg(not(has_fpu))]
+        () => {
+            // Neither `argc` or `argv` make sense in bare metal context so we just
+            // stub them
+            main(0, ::core::ptr::null());
+        }
         #[cfg(has_fpu)]
         () => {
             // NOTE(safe) no exception / interrupt that also accesses the FPU
             // can occur here
             let scb = &*cortex_m::peripheral::SCB.get();
             scb.enable_fpu();
-        }
-        #[cfg(not(has_fpu))]
-        () => {}
-    }
 
-    // Neither `argc` or `argv` make sense in bare metal context so we just
-    // stub them
-    main(0, ::core::ptr::null());
+            // Make sure the user main function never gets inlined into this
+            // function as that may cause FPU related instructions like vpush to
+            // be executed *before* enabling the FPU and that would generate an
+            // exception
+            #[inline(never)]
+            fn main() {
+                unsafe {
+                    ::main(0, ::core::ptr::null());
+                }
+            }
+
+            main()
+        }
+    }
 
     // If `main` returns, then we go into "reactive" mode and simply attend
     // interrupts as they occur.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -155,7 +155,6 @@
 #![feature(used)]
 #![no_std]
 
-#[cfg(feature = "exceptions")]
 extern crate cortex_m;
 extern crate compiler_builtins;
 extern crate r0;
@@ -189,6 +188,18 @@ unsafe extern "C" fn reset_handler() -> ! {
     r0::zero_bss(&mut _sbss, &mut _ebss);
     r0::init_data(&mut _sdata, &mut _edata, &_sidata);
 
+    match () {
+        #[cfg(has_fpu)]
+        () => {
+            // NOTE(safe) no exception / interrupt that also accesses the FPU
+            // can occur here
+            let scb = &*cortex_m::peripheral::SCB.get();
+            scb.enable_fpu();
+        }
+        #[cfg(not(has_fpu))]
+        () => {}
+    }
+
     // Neither `argc` or `argv` make sense in bare metal context so we just
     // stub them
     main(0, ::core::ptr::null());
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -192,6 +191,18 @@ unsafe extern "C" fn reset_handler() -> ! {
     ::r0::zero_bss(&mut _sbss, &mut _ebss);
     ::r0::init_data(&mut _sdata, &mut _edata, &_sidata);
 
+    match () {
+        #[cfg(has_fpu)]
+        () => {
+            // NOTE(safe) no exception / interrupt that also accesses the FPU
+            // can occur here
+            let scb = &*cortex_m::peripheral::SCB.get();
+            scb.enable_fpu();
```
japaric: since main can be inlined into this function, do you think there should have a DSB + ISB right after this?

src/lib.rs
```diff
@@ -192,6 +191,18 @@ unsafe extern "C" fn reset_handler() -> ! {
     ::r0::zero_bss(&mut _sbss, &mut _ebss);
     ::r0::init_data(&mut _sdata, &mut _edata, &_sidata);
 
+    match () {
+        #[cfg(has_fpu)]
+        () => {
+            // NOTE(safe) no exception / interrupt that also accesses the FPU
```
japaric: Do you agree with this? Or should this run within a critical section?

whitequark: Enabling the FPU is always a safe operation.

src/lib.rs
whitequark: No. That's useless. main **must** not be inlined or whichever function it's inlined into will try to save FPU registers and crash before it can do anything (such as enable the FPU).

japaric: We can't prevent LLVM from inlining main from this side though. For instance the following code:

``` rust
fn main() {
    unsafe {
        let x = ptr::read_volatile(0x0 as *const f32);
        let y = ptr::read_volatile(0x0 as *const f32);
        let z = x * y;
        ptr::write_volatile(0x0 as *mut f32, z);
    }
}
```

produces:

``` armasm
08000400 <cortex_m_rt::reset_handler>:
 8000400:	b580      	push	{r7, lr}
 8000402:	466f      	mov	r7, sp
 8000404:	f240 0000 	movw	r0, #0
 8000408:	f240 0100 	movw	r1, #0
 800040c:	f2c2 0000 	movt	r0, #8192	; 0x2000
 8000410:	f2c2 0100 	movt	r1, #8192	; 0x2000
 8000414:	1a09      	subs	r1, r1, r0
 8000416:	f021 0103 	bic.w	r1, r1, #3
 800041a:	f000 f862 	bl	80004e2 <__aeabi_memclr4>
 800041e:	f240 0000 	movw	r0, #0
 8000422:	f240 0100 	movw	r1, #0
 8000426:	f2c2 0000 	movt	r0, #8192	; 0x2000
 800042a:	f2c2 0100 	movt	r1, #8192	; 0x2000
 800042e:	1a09      	subs	r1, r1, r0
 8000430:	f021 0203 	bic.w	r2, r1, #3
 8000434:	f240 41f4 	movw	r1, #1268	; 0x4f4
 8000438:	f6c0 0100 	movt	r1, #2048	; 0x800
 800043c:	f000 f847 	bl	80004ce <__aeabi_memcpy4>
 8000440:	f64e 5088 	movw	r0, #60808	; 0xed88
 8000444:	f2ce 0000 	movt	r0, #57344	; 0xe000
 8000448:	6801      	ldr	r1, [r0, #0]
 800044a:	f441 0170 	orr.w	r1, r1, #15728640	; 0xf00000
 800044e:	6001      	str	r1, [r0, #0]
 8000450:	f240 0000 	movw	r0, #0
 8000454:	f2c0 0000 	movt	r0, #0
 8000458:	7800      	ldrb	r0, [r0, #0]
 800045a:	2000      	movs	r0, #0
 800045c:	ed90 0a00 	vldr	s0, [r0]
 8000460:	ed90 1a00 	vldr	s2, [r0]
 8000464:	ee20 0a01 	vmul.f32	s0, s0, s2
 8000468:	ed80 0a00 	vstr	s0, [r0]
 800046c:	bf30      	wfi
 800046e:	e7fd      	b.n	800046c <cortex_m_rt::reset_handler+0x6c>
```

Only whoever writes main can prevent inlining by adding the #[inline(never)] attribute to it.

whitequark: Sure, just call main through a thunk.
```rust
#[inline(never)]
fn call_main() { main(...); }
call_main();
```

japaric: Oh, that's clever.

japaric:
I'm going to land this so I can test it with #18 but I'm not going to make a release until #18 lands so if anyone has an objection on this PR you can still comment here.

src/lib.rs
```diff
@@ -189,9 +188,34 @@ unsafe extern "C" fn reset_handler() -> ! {
...
+        #[cfg(not(has_fpu))]
+        () => {
+            // Neither `argc` or `argv` make sense in bare metal context so we just
+            // stub them
+            main(0, ::core::ptr::null());
+        }
+        #[cfg(has_fpu)]
+        () => {
+            // NOTE(safe) no exception / interrupt that also accesses the FPU
+            // can occur here
```
whitequark: I feel like this comment is incorrect because enabling the FPU is always safe.

japaric: I probably didn't word this correctly. I'm not referring to the "enable the FPU" operation as being unsafe, but rather that a RMW operation happens here (on FPU related registers but that doesn't matter) and that it's not 100% certain that the RMW operation will be executed atomically because it's *not* executed within a critical section so a data race, by definition, is possible here.

My comment tries to argue that no user defined interrupt / exception handler, which may also access the same memory location (SCB.CPACR), is likely to preempt this operation.

PR #24, opened 2017-07-04
Use the new panic_fmt abi

est31:
Fixes a panic_fmt mismatch with newer rust compiler versions.

This ensures that newer rust compiler versions can optimize out the dead panic formatting machinery and prevents a big bloat of binary sizes.
 
cc https://github.com/rust-lang/rust/issues/43054


japaric:
Thanks @est31.

PR #37, opened 2017-10-11
Put global asm behind target cfg

pftbest:
This will fix 
```
error: <inline asm>:2:1: error: unknown directive
.weak NMI
```
when doing `cargo build` and I hope it will help my IDE a little.


japaric:
Thanks, @pftbest.

PR #38, opened 2017-10-11
Don't include lang_items when in test mode.

pftbest:
This would allow running regular `cargo test` on a crates
that depend on this crate.

Depends on #37


japaric:
Thanks, @pftbest.

PR #41, opened 2017-10-20
[WIP] place static vars at the end of RAM

### LINKED ISSUE #34: (Zero cost) stack overflow protection
Today, the layout of RAM looks like this (assuming no heap and a single RAM region):

today

With this layout when a stack overflow occurs `static` variables end up being overwritten /
corrupted *silently*.

This scenario can be avoided by simply changing the memory layout to look like this:

ideal

In this new scenario a stack overflow will hit the lower RAM boundary. Trying to write beyond the
RAM boundaries raises a HardFault exception. Thus, in theory, the HardFault exception handler
could be used as a stack overflow handler.

In systems where heap memory, which by default starts where the `static` region ends and grows
upwards, exists a similar reordering of the regions can be applied.

### Implementation

To my knowledge this can't be implement using only linker scripts. With a linker script you can
instruct the linker where to *start* a memory region but you can't specify the end address of the
region.

The related bits of the linker script we use are shown below:

```
PROVIDE(_stack_start = ORIGIN(RAM) + LENGTH(RAM));

SECTIONS
{
  /* .. */
  .bss : ALIGN(4)
  {
    _sbss = .;
    *(.bss .bss.*);
    . = ALIGN(4);
    _ebss = .;
  } > RAM

  .data : ALIGN(4)
  {
    _sidata = LOADADDR(.data);
    _sdata = .;
    *(.data .data.*);
    . = ALIGN(4);
    _edata = .;
  } > RAM AT > FLASH

  /* The heap starts right after the .bss + .data section ends */
  _sheap = _edata;

  /* .. */
}
```

A C implementation of this region reordering uses a two step linking process. See 
[this StackOverflow answer](https://stackoverflow.com/a/39477543) for details.

### Other options for stack overflow protection

Assuming that we can't change the memory layout. We could:

- Use the MPU (Memory Protection Unit) to mark the upper boundary of the `static` region as
  read-only. In this scenario when a stack overflow occurs a MemManage exception is raised. This has
  an initialization cost but no runtime cost. The downside is that not all microcontrollers have a
  MPU.

- Implement stack probes for the Cortex-M targets. I believe enabling stack probes carries a runtime
  cost (per function call?) but I'm not sure.

japaric:
so that the stack can never collide into them

closes #34

---

So, this works but there's a downside: the debug information about static
variables is wrong; it says that they are located at the start of the RAM region
when they are actually located at the end of RAM.

I tried @whitequark's suggested approach about placing the static section twice
but ran into two problems:

- ld doesn't like placing the same input section into two different output
sections. Basically if you have something like this

```
._bss : {
  *(.bss);
}

.bss : {
  *(.bss);
}
```

`._bss` will have the input sections but `.bss` will be empty. I think this
makes sense because if duplication was allowed you will end with duplicated
symbols (i.e. the names would crash: there would be a `FOO` in `._bss` and
another `FOO` in `._bss`)

- you can't discard an *output* section. AFAICT, `/DISCARD/` only works on input
sections. 

Assuming that we are OK with losing the debug information about static variables
and landing this we still have to decide how to deal with a heap region. I think
we would have to ask the user for a `_heap_size` to place a fixed size heap
region at the end of RAM, before the static variables. So starting from the end
of RAM the regions would be: heap, static, stack.

cc @pftbest could you or would you be willing to adopt these changes for
msp430-rt?

### DIFF
--- a/link.x
+++ b/link.x
@@ -12,7 +12,7 @@ EXTERN(EXCEPTIONS);
    object file that's passed to the linker *before* this crate */
 EXTERN(INTERRUPTS);
 
-PROVIDE(_stack_start = ORIGIN(RAM) + LENGTH(RAM));
+_eram = ORIGIN(RAM) + LENGTH(RAM);
 
 SECTIONS
 {
@@ -48,22 +48,42 @@ SECTIONS
     . = ALIGN(4);
   } > FLASH
 
+  /* "fake" .bss section, just to get its size */
   .bss : ALIGN(4)
   {
-    _sbss = .;
     *(.bss .bss.*);
     . = ALIGN(4);
-    _ebss = .;
   } > RAM
 
+  /* "fake" .data section, just to get its size */
   .data : ALIGN(4)
   {
-    _sidata = LOADADDR(.data);
-    _sdata = .;
     *(.data .data.*);
     . = ALIGN(4);
-    _edata = .;
   } > RAM AT > FLASH
+
+  /* create the variables that RAM initialization will use */
+  /* here we create the following memory layout */
+  /*                                            */
+  /*  _ebss +-------+ ORIGIN(RAM) + LENGTH(RAM) */
+  /*        | .bss  |                           */
+  /* _edata +-------+ _sbss                     */
+  /*        | .data |                           */
+  /* _sdata +-------+                           */
+  /*        |       |                           */
+  /*        |       |                           */
+  /*        | STACK |                           */
+  /*        |       |                           */
+  /*        |       |                           */
+  /*        +-------+ ORIGIN(RAM)               */
+  _sbss = _eram - SIZEOF(.bss);
+  _ebss = _eram;
+
+  _sdata = _eram - SIZEOF(.bss) - SIZEOF(.data);
+  _edata = _sbss;
+  _sidata = LOADADDR(.data);
+
+  PROVIDE(_stack_start = _sdata);
 
   /* fake output .got section */
   /* Dynamic relocations are unsupported. This section is only used to detect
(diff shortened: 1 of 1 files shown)


pftbest:
Hello, @japaric . I don't think I will port this on MSP430, because chips that I work with have peripheral registers right below the RAM region. So in case of stack overflow program will start writing garbage into peripherals and I think this is even more terrible than overwriting static variables.

japaric:
> So in case of stack overflow program will start writing garbage into peripherals and I think this is even more terrible than overwriting static variables.

That sound horrible.

> the debug information about static variables is wrong

Ugh, I was so focused on getting the start of the stack and the RAM init correct that I failed to notice that not only debug info is wrong but also the operations on the static variables are wrong as well ... the operations are done assuming the static vars are allocated at the start of the RAM region.

So, this actually doesn't work. I'll have to think of some other way.

japaric:
For anyone watching, the next thing that I'm going to try here is the double linking approach. I intend to create a linker wrapper -- let's call it `swap-ld` -- that will invoke `arm-none-eabi-ld` twice. The first time it will invoke it normally. Then `swap-ld` will look at the produced object file to extract the size of the `.bss` and `.data` sections. On the second `arm-none-eabi-ld` invocation `swap-ld` will pass some extra arguments to `arm-none-eabi-ld` to define a few symbols (`_sbss`, `_sdata` and `_sstack` probably). The cortex-m-rt linker script will be tweaked to use these symbols, when defined, to swap the stack and the `.bss+.data` section. The linker script will be modified in such a way that when using vanilla `arm-none-eabi-ld` you get the same memory layout as today but when you use `swap-ld` (e.g. you set it in .cargo/config) as a linker you get the swapped version that provides zero cost stack overflow protection.

I think that approach should work.

japaric:
@whitequark do you have a less complex suggestion?

Note that (a) this is going to be opt-in (you have to specify the `swap-ld` linker to opt in); normal builds will continue to operate the same (single link step and same memory layout as today) and (b) we would control `swap-ld` so we can add all the debugging information we want -- we can even have it enabled by default if that's less surprising (having two link steps is unusual so it might be better to be explicit about it).

I personally think that the extra complexity is worth closing this memory safety hole but I would gladly implement this in a simpler way if I knew how. Regardless of how this is ultimately implemented, I think it's useful to experiment with the alternative memory layout early on to see if there are any other problems with it and to start thinking about how to deal with the heap.

whitequark:
As an optional solution it sounds good, at least in the interim while there is no better option.

japaric:
Closing in favor of #43, which actually works.

PR #43, opened 2017-11-09
Zero cost stack overflow protection, take 2

### LINKED ISSUE #34: (Zero cost) stack overflow protection

japaric:
This is one possible solution to the stack overflow problem described in #34. This approach uses a
linker wrapper, called [swap-ld], to generate the desired memory layout. See #34 for a description
of the desired memory layout and #41 for a description of how `swap-ld` works.

The observable effects of this change in cortex-m programs are:

- the `_sbss` symbol is now override-able.
- there is now a `.stack` linker section that denotes the span of the call stack. `.stack` won't be
  loaded into the program; it just exists for informative purposes (`swap-ld` uses this
  information).

Given the following program:

``` rust
fn main() {
    static mut X: u32 = 0;
    static mut Y: u32 = 1;

    loop {
        unsafe {
            ptr::write_volatile(&mut X, X + 1);
            ptr::write_volatile(&mut Y, Y + 1);
        }
    }
}
```

If you link this program using the `arm-none-eabi-ld` linker, which is the cortex-m-quickstart
default, you'll get the following memory layout:

``` console
$ console
section                                                                   size         addr
.vector_table                                                            0x130    0x8000000
.text                                                                     0x94    0x8000130
.rodata                                                                    0x0    0x80001c4
.stack                                                                  0x5000   0x20000000
.bss                                                                       0x4   0x20000000
.data                                                                      0x4   0x20000004
```

Note how the space reserved for the stack (depicted by the `.stack` linker section) overlaps with
the space where .bss and .data reside.

If you, instead, link this program using `swap-ld` you'll get the following memory layout:

``` console
$ arm-none-eabi-size -Ax app
section                                                                   size         addr
.vector_table                                                            0x130    0x8000000
.text                                                                     0x94    0x8000130
.rodata                                                                    0x0    0x80001c4
.stack                                                                   0x4ff8   0x20000000
.bss                                                                       0x4   0x20004ff8
.data                                                                      0x4   0x20004ffc
```

Note that no overlap exists in this case and that the call stack size has reduced to accommodate the
.bss and .data sections.

Unlike #41 the addresses of static variables is now correct:

``` console
$ arm-none-eabi-objdump -CD app
Disassembly of section .vector_table:

08000000 <_svector_table>:
 8000000:       20004ff8        strdcs  r4, [r0], -r8 ; initial Stack Pointer

08000004 <cortex_m_rt::RESET_VECTOR>:
 8000004:       08000131        stmdaeq r0, {r0, r4, r5, r8}

08000008 <EXCEPTIONS>:
 8000008:       080001bd        stmdaeq r0, {r0, r2, r3, r4, r5, r7, r8}
 (..)

Disassembly of section .stack:

20000000 <.stack>:
        ...

Disassembly of section .bss:

20004ff8 <cortex_m_quickstart::main::X>:
20004ff8:       00000000        andeq   r0, r0, r0

Disassembly of section .data:

20004ffc <_sdata>:
20004ffc:       00000001        andeq   r0, r0, r1
```

closes #34

[swap-ld]: https://github.com/japaric/swap-ld

### TODO

- [x] support a `.heap` region
- [x] update documentation
- [x] publish ~~`swap-ld`~~ `cortex-m-rt-ld`

### DIFF
--- a/link.x
+++ b/link.x
@@ -49,8 +49,12 @@ SECTIONS
   } > FLASH
 
   /* limits of the .stack region */
-  _sstack = ORIGIN(RAM);
   _estack = _stack_start;
+  /* HACK the `true` case indicates that two RAM regions are being used and
+  /* that the stack was placed in the second region. In that case we don't know
+  /* the size of the second RAM region, or its start address, so we just assume
+  /* its zero sized */
+  _sstack = _stack_start < ORIGIN(RAM)? _stack_start : ORIGIN(RAM);
 
   /* fictitious region that represents the memory available for the stack */
   .stack _sstack (INFO) : ALIGN(4)
--- a/link.x
+++ b/link.x
@@ -48,9 +48,19 @@ SECTIONS
     . = ALIGN(4);
   } > FLASH
 
-  .bss : ALIGN(4)
+  /* limits of the .stack region */
+  _sstack = ORIGIN(RAM);
+  _estack = _stack_start;
+
+  /* ficticious region that represents the memory available for the stack */
+  .stack _sstack (INFO) : ALIGN(4)
   {
-    _sbss = .;
+    . += (_estack - _sstack);
+  }
+
+  PROVIDE(_sbss = ORIGIN(RAM));
+  .bss _sbss : ALIGN(4)
+  {
     *(.bss .bss.*);
     . = ALIGN(4);
     _ebss = .;
--- a/link.x
+++ b/link.x
@@ -70,7 +70,7 @@ SECTIONS
     _ebss = .;
   } > RAM
 
-  .data : ALIGN(4)
+  .data _ebss : ALIGN(4)
   {
     _sidata = LOADADDR(.data);
     _sdata = .;
--- a/link.x
+++ b/link.x
@@ -52,7 +52,7 @@ SECTIONS
   _sstack = ORIGIN(RAM);
   _estack = _stack_start;
 
-  /* ficticious region that represents the memory available for the stack */
+  /* fictitious region that represents the memory available for the stack */
   .stack _sstack (INFO) : ALIGN(4)
   {
     . += (_estack - _sstack);
@@ -74,6 +74,17 @@ SECTIONS
     . = ALIGN(4);
     _edata = .;
   } > RAM AT > FLASH
+
+  PROVIDE(_heap_size = 0);
+
+  _sheap = _edata;
+  _eheap = _sheap + _heap_size;
+
+  /* fictitious region that represents the memory available for the heap */
+  .heap _sheap (INFO) : ALIGN(4)
+  {
+    . += _heap_size;
+  }
 
   /* fake output .got section */
   /* Dynamic relocations are unsupported. This section is only used to detect
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -31,7 +31,7 @@
 //!
 //! - A `_sheap` symbol at whose address you can locate a heap.
 //!
-//! - Zero cost stack overflow protection when using the `swap-ld` linker.
+//! - Zero cost stack overflow protection when using the `cortex-m-rt-ld` linker.
 //!
 //! # Example
 //!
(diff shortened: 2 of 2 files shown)


japaric:
Now with `.heap` support.

Without `swap-ld` and with `_heap_size = 0x1000`.

``` console
$ arm-none-eabi-size -Ax app
section                                                                   size         addr
.vector_table                                                            0x400    0x8000000
.text                                                                     0x94    0x8000400
.rodata                                                                    0x0    0x8000494
.stack                                                                  0x5000   0x20000000
.bss                                                                       0x4   0x20000000
.data                                                                      0x4   0x20000004
.heap                                                                   0x1000   0x20000008
```

And with `swap-ld`:

``` console
$ arm-none-eabi-size -Ax app
section                                                                   size         addr
.vector_table                                                            0x400    0x8000000
.text                                                                     0x94    0x8000400
.rodata                                                                    0x0    0x8000494
.stack                                                                   0x3ff8   0x20000000
.bss                                                                       0x4   0x20003ff8
.data                                                                      0x4   0x20003ffc
.heap                                                                   0x1000   0x20004000
```

japaric:
r? @whitequark 

Thoughts on the required changes to implemented this? You expressed concerns about implementing stack overflow protection using this method so I want to know if you find the changes too invasive or complex.

Basically, this PR:

- Made `_sbss` overridable but the symbol is not documented as being overridable as that's an implementation detail of `swap-ld`.
- Added a configurable `_heap_size` symbol, that defaults to 0. This defines the size of the `.heap` section (see below).
- Created ficticious `.stack` and `.heap` sections that represent the RAM regions destined to the stack and the heap. These come in handy when looking at `size -Ax` because they show the size of the stack and the heap. The size of `.stack` is totally wrong (it reports 0) when the stack is placed in a second RAM region, though, because we don't know the name of the region (it could be CCRAM, or RAM2, or something else) so we can't use ORIGIN or LENGTH to figure out its size.

whitequark:
LGTM on the basic concept. I see no better way.

NAK on the name of the linker, just call it `cortex-m-rt-ld` or something like that, `swap-ld` is uninformative at best and may invoke visions of paging at worst. And it's not even swapping anything when the stack is in another memory region.

japaric:
@whitequark Thanks for the feedback. Agree on renaming the linker wrapper.

PR #50, opened 2017-12-30
[RFC] enforce `main: fn() -> !`

japaric:
Now that the `Termination` trait has been implemented (cf. rust-lang/rust#46479) we can narrow down
the return type of `main` to be one of the types that implements the `Termination` trait.

This RFC proposes *only* implementing `Termination` for the never type (`!`) -- currently
`Termination` is only implemented for the unit type (`()`). With this change all programs that link
to `cortex-m-rt` are forced to set the signature of `main` to be `fn() -> !`.  The rationale here is
that a divergent (never ending) function better matches the logic of microcontroller programs.

When the change is implemented the user will be forced to make `main` into a divergent function.

This won't work:

``` rust
#![feature(termination_trait)]
#![no_std]

fn main() {}
//~^ error: the trait `cortex_m_rt::lang_items::Termination` is not implemented for `()`
```

Whereas this would works:

``` rust
#![feature(termination_trait)]
#![no_std]

fn main() -> ! {
    loop {}
}
```

### Drawbacks

Diagnostics for the `Termination` feature are rather bad at the moment. If this change is
implemented and you write a program where `main: fn()` but forget to add the
`#![feature(termination_trait)]` feature gate you get an unhelpful *linker* error:

``` rust
#![no_std]

fn main() {}
```

``` console
$ xargo build
(..)
  = note: /home/japaric/tmp/cortex-m-quickstart/target/thumbv7m-none-eabi/debug/deps/cortex_m_quickstart-0f691e4b3fe1bbf7.cortex_m_quickstart.rcgu.o: In function `main':
          cortex_m_quickstart-fd49a734b68c9115d8712982b4fdc764.rs:(.text.main+0x36): undefined reference to `cortex_m_rt::lang_items::start'
```

cc @pftbest

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ keywords = ["arm", "cortex-m", "runtime", "startup"]
 license = "MIT OR Apache-2.0"
 name = "cortex-m-rt"
 repository = "https://github.com/japaric/cortex-m-rt"
-version = "0.3.8"
+version = "0.4.0"
 
 [dependencies]
 cortex-m = "0.3.0"
--- a/src/lang_items.rs
+++ b/src/lang_items.rs
@@ -38,7 +38,7 @@ pub trait Termination {
     fn report(self) -> i32;
 }
 
-impl Termination for () {
+impl Termination for ! {
     fn report(self) -> i32 {
         0
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -287,6 +287,7 @@
 #![feature(lang_items)]
 #![feature(linkage)]
 #![feature(naked_functions)]
+#![feature(never_type)]
 #![feature(used)]
 #![no_std]
 
@@ -307,8 +308,10 @@ use cortex_m::exception::ExceptionFrame;
 
 extern "C" {
     // NOTE `rustc` forces this signature on us. See `src/lang_items.rs`
+    // NOTE the return type can only be the never type (`!`) because that's the only type that
+    // implements the `Termination` trait
     #[cfg(target_arch = "arm")]
-    fn main(argc: isize, argv: *const *const u8) -> isize;
+    fn main(argc: isize, argv: *const *const u8) -> !;
 
     // Boundaries of the .bss section
     static mut _ebss: u32;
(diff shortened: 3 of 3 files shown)


pftbest:
This change seems logical, and I like it, but I'm not very happy about adding one more `#![feature]` to the user's code. It's one more roadblock on the path to stabilization. But maybe it will be stable soon, because it's a very popular feature.

Also I found one more error message that is not very helpful:
```Rust
fn main() -> ! {
    while true {
    }
}
```
```
error[E0308]: mismatched types
 --> <source>:2:5
  |
1 |   fn main() -> ! {
  |                - expected `!` because of return type
2 | /     while true {
3 | |     }
  | |_____^ expected !, found ()
  |
  = note: expected type `!`
             found type `()`
```

japaric:
> I'm not very happy about adding one more #![feature] to the user's code.

Me neither. I wish the diagnostics for forgetting the feature gate were better. We can always wait until the feature gets stabilized.

> Also I found one more error message that is not very helpful:

That seems unrelated to the termination trait. You get the same error with any other function:

``` rust
fn foo() -> ! {
    while true {}
    //^ error: expected !, found ()
}
```

I think the compiler doesn't check the `while` condition to pick the return type accordingly. That may be related to const eval; for instance should this one work (be interpreted as `!`): `let cond = true; while cond {}`? what about this one: `const COND: bool = true; while COND {}`?

pftbest:
@japaric  sorry, I didn't mean to say that `while cond {}` should be evaluated at compile time, I just noticed a missing warning.

When you write this code:
```Rust
fn main() {
    while true {
    }
}
```
You get a nice warning that suggests using `loop` instead:
```
warning: denote infinite loops with `loop { ... }`
 --> <source>:3:5
  |
3 |     while true {
  |     ^^^^^^^^^^ help: use `loop`
  |
  = note: #[warn(while_true)] on by default
```

But when you have a diverging function, you don't see such message.

japaric:
Oh, OK. That sounds like it might be worth to report in rust-lang/rust?

japaric:
This has been done. The `entry!` requires the entry point to be a divergent function.

PR #57, opened 2018-02-16
Support dev channels of rustc

crawford:
Official channels have a commit_date, but local builds do not, causing
the build script to panic.


japaric:
Thanks @crawford

PR #64, opened 2018-04-06
v0.4.0: with LLD support

japaric:
See commit messages for details about why this is a breaking change.

I personally think no longer requiring GNU LD / arm-none-eabi-binutils to build a Rust program is worth the breaking change.

So far this seems to work fine but I'll continue to test this locally.

This depends on rust-lang/rust#49728

### DIFF
--- a/link.x
+++ b/link.x
@@ -12,7 +12,7 @@ EXTERN(EXCEPTIONS);
    object file that's passed to the linker *before* this crate */
 EXTERN(INTERRUPTS);
 
-_stack_start = ORIGIN(RAM) + LENGTH(RAM);
+PROVIDE(_stack_start = ORIGIN(RAM) + LENGTH(RAM));
 
 SECTIONS
 {
--- a/link.x
+++ b/link.x
@@ -68,11 +68,8 @@ SECTIONS
     _edata = .;
   } > RAM AT > FLASH
 
-  PROVIDE(_heap_size = 0);
-
   /* The heap starts right after the .bss + .data section ends */
   _sheap = _edata;
-  _eheap = _sheap + _heap_size;
 
   /* fake output .got section */
   /* Dynamic relocations are unsupported. This section is only used to detect
--- a/link.x
+++ b/link.x
@@ -54,7 +54,10 @@ SECTIONS
     *(.bss .bss.*);
     . = ALIGN(4);
     _ebss = .;
-  } > RAM
+  } > RAM AT > FLASH
+  /* NOTE(AT > FLASH) without this LLD v6 produces a binary that crashes OpenOCD whereas LLD v7
+     emits a ".rodata and .bss sections overlap" error ... This hacky workaround doesn't increase
+     the binary size AFAICT */
 
   .data : ALIGN(4)
   {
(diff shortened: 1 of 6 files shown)


whitequark:
Dropping stack overflow protection is *kind of* unfortunate but given that it required the custom linking step maybe it's good that it's gone. I think that a proper solution can be much more easily developed for lld than binutils, especially given that rustc ships a fork of lld now.

link.x
```diff
@@ -54,7 +54,10 @@ SECTIONS
     *(.bss .bss.*);
     . = ALIGN(4);
     _ebss = .;
-  } > RAM
+  } > RAM AT > FLASH
+  /* NOTE(AT > FLASH) without this LLD v6 produces a binary that crashes OpenOCD whereas LLD v7
+     emits a ".rodata and .bss sections overlap" error ... This hacky workaround doesn't increase
+     the binary size AFAICT */
```
whitequark: One drawback is this will make firmwares with *really* full flashes uncompilable. Do I understand it right that lld v7 doesn't require the workaround?

japaric: > Do I understand it right that lld v7 doesn't require the workaround?

No, this also needed for the v7 I tested. Without the workaround linking with LLD fails with the error message in the comment; GNU LD can link the same program without errors and produces a binary that doesn't crash OpenOCD.

whitequark: Are you going to look at the underlying lld bug? I don't yet use lld for my embedded development so I'm afraid you're in the better position here. I can review though.

japaric: I will at least submit a bug report. I have zero experience hacking LLVM :-).

therealprof:
I just tested this with my microbit crate. Works fine with both arm-none-eabi-binutils and lld. Flash image size is identical, disk size with debugging symbols went down a bit and surprisingly lld produces slightly larger files on disk than GNU ld but that's nothing to worry about.

japaric:
@whitequark 

> I think that a proper solution can be much more easily developed for lld than binutils, especially given that rustc ships a fork of lld now.

I think it's very unlikely that the Rust team will add features to rust fork of lld. They try to keep as few patches (that are not backports) as possible on top of upstream LLVM / LLD. I do agree that a LLD (linker script?) solution would be the ideal solution here.

japaric:
As we are making a new version release I'm also removing the panic_fmt lang item and the abort-on-panic feature. rust-lang/rust#47074 has been fixed and `#[panic_implementation]` is around the corner so it seems a good time to remove them. Without them the user is expected to specify the panic behavior; they can do so using `#[panic_implementation]` or by depending on a [panic crate](https://github.com/japaric/panic-abort).

whitequark:
> I think it's very unlikely that the Rust team will add features to rust fork of lld. They try to keep as few patches (that are not backports) as possible on top of upstream LLVM / LLD. I do agree that a LLD (linker script?) solution would be the ideal solution here.

I mean that the features can be added upstream and then pulled to the rust fork of lld to avoid waiting for an LLVM release cycle, of course.

japaric:
Alright, let's land this to get more real world testing.

PR #66, opened 2018-04-18
Use linker script to create weak aliases for exception handlers.

pftbest:
Fixes https://github.com/japaric/cortex-m-rt/issues/58

This is a breaking change for projects that are using custom linker
scripts.


japaric:
Done in #69

PR #67, opened 2018-04-21
Fix typo in Exception name

njgrant:
This will make the name in `enum Exception` match the name of the exception handler function, like they already do for all other exceptions.


japaric:
Thanks for the PR. This exception was renamed in #69.

PR #68, opened 2018-04-24
Update cortex-m dependency

burrbull:


japaric:
As of v0.5.0 this crate doesn't depend on `cortex-m`.

Issue #53, opened 2018-01-17
LLD support (AKA LLD *almost* works out of the box)


japaric:
If you switch the linker from arm-none-eabi-ld to ld.lld you'll get the following error:

``` console
$ ld.lld --version
LLD 7.0.0 (https://github.com/llvm-mirror/lld 527f7fd20d23a54ae30ccdf5071312b24e28776b) (compatible with GNU linkers)

$ xargo build
error: linking with `ld.lld` failed: exit code: 1
  |
  = note: "ld.lld" (..)
  = note: ld.lld: error: $PWD/target/thumbv7em-none-eabihf/debug/build/cortex-m-rt-f07298b7bbb47f59/out/link.x:92: NOLOAD expected, but got INFO
          >>>   .debug_gdb_scripts _stext (INFO) : {
          >>>                              ^
```

But if you modify this crate link.x like this:

``` diff
   /* a rustc hack will force the program to read the first byte of this section,
      so we'll set the (fake) start address of this section to something we're
      sure can be read at runtime: the start of the .text section */
-  .debug_gdb_scripts _stext (INFO) : {
+  .debug_gdb_scripts _stext (NOLOAD) : {
     KEEP(*(.debug_gdb_scripts))
-  }
+  } > FLASH
```

Programs link correctly. The resulting binaries are well formed and debug information works. :tada:

However, this results in an *artifical* increase of size of the text section as reported by `arm-none-eabi-size`:

``` console
$ # before (linked with GNU ld)
$ arm-none-eabi-size (..)
   text    data     bss     dec     hex filename
   1300       0       0    1300     514 (..)

$ arm-none-eabi-size -Ax (..)
section                  size         addr
.vector_table           0x188    0x8000000
.text                   0x384    0x8000188
.rodata                   0x8    0x800050c
.bss                      0x0   0x20000000
.data                     0x0   0x20000000
.debug_gdb_scripts       0x88    0x8000188

$ # after (linked with GNU ld)
$ arm-none-eabi-size (..)
   text    data     bss     dec     hex filename
   1436       0       0    1436     59c (..)

$ arm-none-eabi-size -Ax (..)
section                  size         addr
.vector_table           0x188    0x8000000
.text                   0x384    0x8000188
.rodata                   0x8    0x800050c
.bss                      0x0   0x20000000
.data                     0x0   0x20000000
.debug_gdb_scripts       0x88    0x8000188 <- NOTE overlaps with .text

[...]

PR #69, opened 2018-04-24
compile on stable

japaric:
with these changes this crate compiles on stable


japaric:
It also appears to fix all the LLD issues reported in #53. cc @jcsoo

src/lib.rs
```diff
@@ -1,615 +1,210 @@
...
-DEBUG_MONITOR = DEFAULT_HANDLER
-"#
-);
+/* Exceptions */
+// NOTE we purposefully go against Rust style here and use PascalCase for the handlers. The
+// rationale is in the definition of the `exception!` macro.
+#[doc(hidden)]
+pub enum Exception {
+    NMI,
+    MenManage,
```
adamgreig: This should probably be Me*m*Manage?

jcsoo:
Tried it out and it works for me.

japaric:
Changes from cortex-m-rt v0.4.x and stable-embedded-rust that warrant some discussion:

- The user must define *both* `DefaultHandler` and `HardFault`.

- The `HardFault` handler is now a diverging function (`fn(..) -> !`). The rationale is that
  returning from the hard fault handler can be (is usually?) UB. A diverging function prevents that
  scenario.

- Now, only the `HardFault` handler has access to the previous stack frame. In `cortex-m-rt`
  `DefaultHandler` had access to the previous stack frame.

- `DefaultHandler` now takes a `u8` argument. This integer is the exception number and can be used
  to identify which exception / interrupt is being serviced.

- `interrupts!` can be used to bind all the potential interrupt handlers (240) to `DefaultHandler`.
  This can be used to write device agnostic programs. In `cortex-m-rt` one had to explicitly declare
  a `static` variable and place it in the `.vector_table.interrupts` section; `interrupts!` is (less
  error prone) sugar for that.

Unresolved questions:

- Do we want to apply the divergent function trick to the other *fault* handlers; that is to
  `UsageFault`, `BusFault` and `MemManage` (MPU fault?)? Is it safe (not UB) to return from those
  fault handlers? An idea here could be to make it safe to use a divergent function but unsafe to
  use a non divergent function. See below:

``` rust
// You can pick
exception!(BusFault, bf);

fn bf() -> ! {
    // ..
}

// OR this
exception!(unsafe BusFault, bf);

fn bf()  {
    // ..
    return;
}
```

- Do we want one default handler for Cortex-M exceptions and a different default handler for
  interrupts? That is `DefaultExceptionHandler` and `DefaultInterruptHandler`. (I don't see much
  gain from such split)

TODO:

- [ ] Add back exception / interrupt local data (`locals` in the old `exception!` macro)

- [ ] Document the overridable symbols (`_foo`) and the private symbols (`__foo`). The later will
  be documented for inspection / diagnostic purposes (`objdump`, `nm`, etc.).

- [ ] Document that `exception!` and `interrupts!` must be called in the root of the crate, and if possible add (reachability) checks for that.

cc @thejpster

japaric:
Bikeshedding: should `main!` be renamed to `entry!`? The macro defines the user program entry point, which doesn't need to be named `main`.

japaric:
Future improvements:

Macros 1.2 is planned for the 1.28 release and will allow custom attributes (i.e. `#[foo]`; today we only allow custom derive on stable: `#[derive(Foo)]`), so we could in the future replace the `exception!` macro for an `#[exception]` attribute which would slightly reduce boilerplate / look more natural:

``` rust
// Today
#[macro_use(exception)]
extern crate cortex_m_rt;

exception!(SysTick, sys_tick);

fn sys_tick() { .. }

// Macros 1.2
extern crate cortex_m_rt;

use cortex_m_rt::exception;

#[exception(SysTick)]
fn sys_tick() { .. }

// or even
#[exception]
fn SysTick() { .. }
```

We could do the same with `main!` macro; we could replace it with an `#[entry]` attribute.

jcsoo:
> Do we want one default handler for Cortex-M exceptions and a different default handler for
interrupts? That is DefaultExceptionHandler and DefaultInterruptHandler. (I don't see much
gain from such split)

In my use case, I want all interrupts to optionally go to a shared handler, but I'd prefer to have individual exception handlers as well as a DefaultHandler for everything else.

Is it possible to have the interrupts! macro bind to a specific handler rather than DefaultHandler?

japaric:
> Is it possible to have the interrupts! macro bind to a specific handler rather than DefaultHandler?

It could but most applications are *not* going to use the `interrupts!` macro; they are going to use the `__INTERRUPTS` variable provided by a device crate (generated using svd2rust) or an equivalent to it.

So, I guess my question actually is: should `svd2rust` bind the device specific interrupts to `DefaultHandler` or to some other handler (`DefaultInterruptHandler`)? Though if we go down that route we'll probably want to add `EXTERN(DefaultInterruptHandler)` to `link.x` to prevent the linker from dropping the symbol.

If you are using something other than svd2rust you can do whatever you want; you don't need to use `interrupts!` (`svd2rust` doesn't use `interrupts!` either)

japaric:
bors try

PR #78, opened 2018-06-13
Modified the entry and exception macros to accept a closure.

sjroe:
This allows:

```rust
entry!(|| {
    let mut x = 1;

    loop {
        x = x + 1;
    }
});
```

as well as allowing the original usage:

```rust
entry!(main);
fn main() -> ! {
    let mut x = 1;

    loop {
        x = x + 1;
    }
}
```

The same is true for exceptions:

```rust
exception!(*, |irqn: i16| {
    panic!("Unhandled exception (IRQn = {})", irqn);
});
```


japaric:
Nice! Thanks @sjroe 

bors r+

PR #81, opened 2018-07-20
Add a __pre_init function to be called at the start of the reset handler

### LINKED ISSUE #17: add some unsafe method to run code before RAM is initialized
Some people have expressed that they would like to configure the CPU clock before RAM initialization to speed up the initialization process (specially when lots of RAM need to be initialized).

This has to be unsafe because all the static variables stored in .bss / .data would be uninitialized during the execution of this pre-init routine.

Possible user facing interface:

``` rust
#[macro_use]
extern crate cortex_m_rt;

pre_init!(unsafe {
    // something
});
```

where `pre_init!` can only be used once.

yodaldevoid:
I needed this for a project so I went ahead and rebased and tweaked @jcsoo 's changes from #71. I will admit, I got a little impatient waiting and that also played into why I did this. If either @jcsoo or @japaric have an issue with this PR, please let me know and I will remove it. I apologize for toes that I have stepped on.

Now onto the PR. This is possibly the simplest implementation that is possible. No nightly features were used in this implementation. Also, there is no hand-holding for the end user; if they want to mess with uninitialized statics, that is on them.

If you would like me to add more documentation, please let me know what you would like to see.

Fixes #17

### DIFF
--- a/examples/pre_init.rs
+++ b/examples/pre_init.rs
@@ -4,11 +4,9 @@
 #![no_main]
 #![no_std]
 
-#[macro_use(entry, exception, pre_init)]
+#[macro_use(entry, pre_init)]
 extern crate cortex_m_rt as rt;
 extern crate panic_semihosting;
-
-use rt::ExceptionFrame;
 
 pre_init!(disable_watchdog);
 
@@ -22,15 +20,3 @@ entry!(main);
 fn main() -> ! {
     loop {}
 }
-
-// the hard fault handler
-exception!(HardFault, hard_fault);
-
-fn hard_fault(_ef: &ExceptionFrame) -> ! {
-    loop {}
-}
-
-// the default exception handler
-exception!(*, default_handler);
-
-fn default_handler(_irqn: i16) {}
--- a/examples/pre_init.rs
+++ b/examples/pre_init.rs
@@ -1,0 +1,36 @@
+//! `cortex-m-rt` based program with a function run before RAM is initialized.
+
+#![deny(warnings)]
+#![no_main]
+#![no_std]
+
+#[macro_use(entry, exception, pre_init)]
+extern crate cortex_m_rt as rt;
+extern crate panic_semihosting;
+
+use rt::ExceptionFrame;
+
+pre_init!(disable_watchdog);
+
+unsafe fn disable_watchdog() {
+    // Do what you need to disable the watchdog.
+}
+
+// the program entry point
+entry!(main);
+
+fn main() -> ! {
+    loop {}
+}
+
+// the hard fault handler
+exception!(HardFault, hard_fault);
+
+fn hard_fault(_ef: &ExceptionFrame) -> ! {
+    loop {}
+}
+
+// the default exception handler
+exception!(*, default_handler);
+
+fn default_handler(_irqn: i16) {}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -169,13 +169,6 @@
 //! conjunction with crates generated using `svd2rust`. Those *device crates* will populate the
 //! missing part of the vector table when their `"rt"` feature is enabled.
 //!
-//! ## `pre_init`
-//!
-//! If this feature is enabled then a user-defined function will be run at the start of the reset
-//! handler, before RAM is initialized. If this feature is enabled then the macro `pre_init!` needs
-//! to be called to set the function to be run. This feature is intended to perform actions that
-//! cannot wait the time it takes for RAM to be initialized, such as disabling a watchdog.
-//!
 //! # Inspection
 //!
 //! This section covers how to inspect a binary that builds on top of `cortex-m-rt`.
@@ -243,6 +236,10 @@
 //! size depends on the target device but if the `"device"` feature has not been enabled it will
 //! have a size of 32 vectors (on ARMv6-M) or 240 vectors (on ARMv7-M). This array is located after
 //! `__EXCEPTIONS` in the `.vector_table` section.
+//!
+//! - `__pre_init`. This is a function to be run before RAM is initialized. It defaults pointing at
+//! `0` and if not changed to point to another address, usually by calling the `pre_init!` macro,
+//! the `_pre_init` function is skipped.
 //!
 //! If you override any exception handler you'll find it as an unmangled symbol, e.g. `SysTick` or
 //! `SVCall`, in the output of `objdump`,
@@ -385,6 +382,14 @@
 //!     println!("cargo:rustc-link-search={}", out.display());
 //! }
 //! ```
+//!
+//! ## `pre_init!`
+//!
+//! A user-defined function can be run at the start of the reset handler, before RAM is
+//! initialized. The macro `pre_init!` can be called to set the function to be run. The function is
+//! intended to perform actions that cannot wait the time it takes for RAM to be initialized, such
+//! as disabling a watchdog. As the function is called before RAM is initialized, any access of
+//! static variables will result in undefined behavior.
 
 // # Developer notes
 //
@@ -482,12 +487,13 @@ pub unsafe extern "C" fn Reset() -> ! {
         static mut __edata: u32;
         static __sidata: u32;
 
-        #[cfg(feature = "pre_init")]
-        fn _pre_init();
+        fn __pre_init();
     }
 
-    #[cfg(feature = "pre_init")]
-    _pre_init();
+    let pre_init: unsafe extern "C" fn() = __pre_init;
+    if pre_init as usize != 0 {
+        pre_init();
+    }
 
     // Initialize RAM
     r0::zero_bss(&mut __sbss, &mut __ebss);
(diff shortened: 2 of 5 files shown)


korken89:
Hi @japaric, could you give this a look? Cf. #17

yodaldevoid:
@japaric Pinging to make sure this didn't get lost in the shuffle.

japaric:
@yodaldevoid thanks for changes. I was looking at the disassembly and it seems that the (LLVM?) optimization passes assume that all function pointers can't be null so __pre_init is *always* called. We'll have to choose a different sentinel value.

Could you also add an example for this to he `examples` directory?

> it would be good if someone can test that this doesn't break existing code when pre_init! isn't used.

The examples are compiled as part of the test suite so you can be sure that no accepted change will break linking those examples.

korken89:
@yodaldevoid:

I did some testing, setting the sentinel value to 1 instead of 0 fixes the problem and the comparing instructions are added.

Before (0):
```
08000424 <Reset>:
 8000424:	f000 f840 	bl	80004a8 <____pre_init_veneer>
```

New (1):
```
08000424 <Reset>:
 8000424:	f240 0001 	movw	r0, #1
 8000428:	f2c0 0000 	movt	r0, #0
 800042c:	2801      	cmp	r0, #1
 800042e:	bf18      	it	ne
 8000430:	f000 f83e 	blne	80004b0 <____pre_init_veneer>
```

With `pre_init!` (looks strange due to some `#[inline(never)]` sprinkling):
```
0800042e <Reset>:
 800042e:	f240 4011 	movw	r0, #1041	; 0x411
 8000432:	f6c0 0000 	movt	r0, #2048	; 0x800
 8000436:	2801      	cmp	r0, #1
 8000438:	d135      	bne.n	80004a6 <Reset+0x78>

...

 80004a6:	f7ff ffb3 	bl	8000410 <__pre_init>

...

08000400 <_ZN20test7my_init17hd14b71b0b4782276E>:
 8000400:	be00      	bkpt	0x0000

...

08000410 <__pre_init>:
 8000410:	f7ff fff6 	bl	8000400 <_ZN20test7my_init17hd14b71b0b4782276E>

```

Could you fix this and add the example @japaric asked about? I'd like to get this PR moving. :)

@japaric:

> A more general approach would be implement .init_array then any crate would able to inject code that will be run before main but I just checked and that requires the unstable #[used] attribute on the caller site or the thing doesn't work when optimizations are applied. (OTOH, letting any crate run code before main sounds hard to reason about.

Indeed, I would not do this. When running code before the init of RAM it is commonly of paramount importance to know the order of operations, so merging it into `.init_array` can cause unexpected behavior, together with `.init_array` being performed after RAM init. I would keep it as a single `pre_init` function as it is here.

yodaldevoid:
@japaric 
> The examples are compiled as part of the test suite so you can be sure that no accepted change will break linking those examples.

In terms of people checking if this was broken or not, I was asking for someone to do the exact check that you did. Checking the disassembled binary completely slipped my mind.

I will be sure to add an example.

@korken89 
> I did some testing, setting the sentinel value to 1 instead of 0 fixes the problem and the comparing instructions are added.

I figured this would work, but some problems came up this week that kept me from working on anything this week. I will try to get everything taken care of either tonight or tomorrow.

yodaldevoid:
@japaric I have added an example and changed the sentinel value to 1. Looking at the disassembly it seems to be working. The branch is optimized out when a function is provided, though not when none is provided.

src/lib.rs
```diff
@@ -237,6 +237,10 @@
 //! have a size of 32 vectors (on ARMv6-M) or 240 vectors (on ARMv7-M). This array is located after
 //! `__EXCEPTIONS` in the `.vector_table` section.
 //!
+//! - `__pre_init`. This is a function to be run before RAM is initialized. It defaults pointing at
+//! `0` and if not changed to point to another address, usually by calling the `pre_init!` macro,
```
korken89: There is an error here, should be 1

adamgreig:
Please could you add the pre_init example to `ci/script.sh`?

examples/pre_init.rs
```diff
@@ -0,0 +1,36 @@
...
+}
+
+// the program entry point
+entry!(main);
+
+fn main() -> ! {
+    loop {}
+}
+
+// the hard fault handler
```
korken89: The `exception!` handlers here are not needed, as they are defaulted now.

yodaldevoid: Ok, I can remove them. I simply copied and changed the code from the "minimal" example without looking too closely at everything else.

src/lib.rs
yodaldevoid: Ah, I knew I should have looked at the docs once again!

adamgreig: It would be useful to include a short comment about why it can't be 0 too.

adamgreig:
> The branch is optimized out when a function is provided, though not when none is provided.

I'm not seeing this optimisation kick in. If I build the pre_init and alignment examples in `--release` I still get the check and branch right at the start:

`pre_init` example, showing branch to `_stext` which contains optimised `__pre_init`:
pre_init_start_pr81

`alignment` example, as per master, showing no comparison or branch:
alignment_master

`alignment` example, with this PR, showing we still load and compare:
alignment_start_pr81

Is this the expected behaviour? Can we optimise out this check when the branch will never be taken?

korken89:
I don't think this can be optimised as the required address is added in the linking stage, I have seen this behavior during all testing of these additions with the sentinel value.

Though I think @japaric can abet this better.

adamgreig:
In that case maybe we should actually provide a `DefaultPreInit` which is an empty function, `PROVIDE` that in the linker script, remove the check and always call `__pre_init` in `Reset`? It emits fewer instructions in both cases (with and without a user function), as far as I can tell.

adamgreig:
With the empty `DefaultPreInit` suggested above, in an example with no `pre_init!` call, built with release and fat LTO:

```
08000438 <Reset>:
 8000438:	f000 f933 	bl	80006a2 <DefaultPreInit>
 800043c:	f240 0008 	movw	r0, #8
 8000440:	f240 0204 	movw	r2, #4
 8000444:	f2c2 0000 	movt	r0, #8192	; 0x2000
 8000448:	f2c2 0200 	movt	r2, #8192	; 0x2000
 800044c:	4282      	cmp	r2, r0
 800044e:	d25a      	bcs.n	8000506 <Reset+0xce>
...
080006a2 <DefaultPreInit>:
 80006a2:	4770      	bx	lr
```

Note the branch to `DefaultPreInit` which immediately returns. Two instructions/six bytes/two branch overhead.

Example with `pre_init!()` pointing to an empty function (the current `pre_init` example):
```
08000400 <Reset>:
 8000400:	f240 0000 	movw	r0, #0
 8000404:	f240 0200 	movw	r2, #0
 8000408:	f2c2 0000 	movt	r0, #8192	; 0x2000
 800040c:	f2c2 0200 	movt	r2, #8192	; 0x2000
 8000410:	4282      	cmp	r2, r0
 8000412:	d25a      	bcs.n	80004ca <Reset+0xca>
```
DefaultPreInit is omitted entirely, no overhead, same as we currently have on master.

Example with `pre_init!` doing a `ptr::write_volatile(0xcafe as *mut u32, 0xbeef)`:

```
08000400 <Reset>:
 8000400:	f64c 20fe 	movw	r0, #51966	; 0xcafe
 8000404:	f64b 61ef 	movw	r1, #48879	; 0xbeef
 8000408:	f240 0200 	movw	r2, #0
 800040c:	6001      	str	r1, [r0, #0]
 800040e:	f240 0000 	movw	r0, #0
 8000412:	f2c2 0200 	movt	r2, #8192	; 0x2000
 8000416:	f2c2 0000 	movt	r0, #8192	; 0x2000
 800041a:	4282      	cmp	r2, r0
 800041c:	d25a      	bcs.n	80004d4 <Reset+0xd4>
```

The new (and inlined) function is the very first thing to run. No other overhead.

src/lib.rs
```diff
@@ -872,3 +893,35 @@ macro_rules! exception {
...
+///
+/// mod foo {
+///     pub unsafe fn bar() {
+///         // do something here
+///     }
+/// }
+/// ```
+#[macro_export]
+macro_rules! pre_init {
+    ($handler:path) => {
```
adamgreig: @yodaldevoid sorry to add yet another thing! But please could you change `path` to `expr` here, as per the entry/exception macros, so that it can accept a closure as well as a function path? The docstrings could then be updated accordingly too.

This changed in #78 for the other macros.

yodaldevoid: So, I tried this when I was originally writing this as I saw that PR before it was merged, but this does not work as you cannot make an `unsafe` closure and the function passed needs to be unsafe as stated before by @japaric [here](https://github.com/rust-embedded/cortex-m-rt/pull/71#issuecomment-388630004).

adamgreig: Ah, that's a shame. Fair enough.

yodaldevoid:
@adamgreig I tried making a simple empty default pre-init function and built with fat LTO and didn't see it inlined. That said, I like that implementation better if only to get rid of the conditional branch.

adamgreig:
Right, if you have a empty default pre-init it doesn't get inlined (my first example) but is at least a lot less overhead than the conditional approach. If you then use `pre_init!(some_empty_function)` it's entirely optimised out (my second example), so for people who were concerned about the small startup overhead, there's a very simple way to get rid of it entirely.

yodaldevoid:
I've changed over to using the empty default function instead of a sentinel as suggested by @adamgreig. Please feel free to rip it to shreds.

adamgreig:
Thanks! LGTM. I'll let @korken89 approve his review and then I think we're good to merge, maybe we can sneak it in to v0.5.2 #88 too.

yodaldevoid:
Looks like one of the travis builds stalled.

korken89:
This does indeed generate better assembly, good catch and good work!

bors r+

PR #85, opened 2018-08-08
[WIP] provide defaults for DefaultHandler and HardFault

### LINKED ISSUE #72: RFC: Provide defaults for `HardFault` and `DefaultHandler`?
Currently these two symbols must be defined somewhere in the dependency graph of an application --
though they are usually defined in the application crate itself.

We could provide defaults for these two symbols in this crate obviating the need to define them in
some cases. The questions to answer are: should we? and if we do what should the defaults be?

Some thoughts to kick off the discussion

# Should we do this?

There's a trade-off between ergonomics for the experienced used (as in knows what exceptions are
and what the HardFault exception means) and learnability for people getting started on embedded
development.

For the experienced user defaults are good because that just means "less boilerplate" for trivial
applications, examples and programs that are being debugged; and they know when they should override
these exception handlers: non trivial applications, production builds, etc.

On the other hand, defaults may make the beginner never learn about these handlers, think that they
are not important and/or think that the defaults do the right thing in every possible situation ("sane
defaults"). This can turn into a nasty surprise down road when they run their application detached
from the debugger and end up experiencing a crash / lockup with no means of investigating the cause
because they didn't override the handlers to log faults.

# What should the defaults be?

Likely bad options:

- Hardware breakpoint (`bkpt`). `bkpt` triggers a `HardFault` exception. When used from the
  `HardFault` handler it results in infinite recursion.

- `intrinsics::abort`. `abort` triggers a `HardFault` exception so same problem as above.

- `panic!`. The semantics of `panic!` varies per application making it not general enough to be used
  here. For example, `panic!` could trigger a hardware breakpoint or call `abort`.

What others do:

- CMSIS sets `DefaultHandler` (and `HardFault`) to an infinite loop.

Other considerations:

- `DefaultHandler` and `UserHardFault` are stable unmangled symbols that one can place breakpoints
  at. So if you have a debugger attached to the device you can always catch hard fault and unhandled
  exceptions before the handlers are executed.

cc @therealprof @jcsoo @thejpster

japaric:
`exception!(HardFault, ..)` and `exception!(*, ..)` can now be omitted from
programs to pick up the default behavior of an infinite loop. Existing programs
that define these exception handlers will continue to work w/o any functional
change.

closes #72

--

Annoyances:

- The handlers can't be *just* an infinite loop because of rust-lang/rust#28728.
If we define the handlers as just `loop {}` they will become an abort (UDF)
instruction. And that would turn HardFault into infinite recursion. For that
reason I have made them into an infinite loop that does some side effect

- If you stick to the defaults then the symbol name of the default handler
changes from `DefaultHandler` (override) to `DefaultDefaultHandler` (default).
We can make these two names more similar but I think we can not prevent the
rename. Something similar happens with UserHardFault (which becomes
DefaultUserHardFault when not overridden).

cc @rust-embedded/cortex-m

### DIFF
--- a/link.x.in
+++ b/link.x.in
@@ -45,6 +45,9 @@ PROVIDE(DebugMonitor = DefaultHandler);
 PROVIDE(PendSV = DefaultHandler);
 PROVIDE(SysTick = DefaultHandler);
 
+PROVIDE(DefaultHandler = DefaultDefaultHandler);
+PROVIDE(UserHardFault = DefaultUserHardFault);
+
 /* # Interrupt vectors */
 EXTERN(__INTERRUPTS); /* `static` variable similar to `__EXCEPTIONS` */
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -390,7 +390,7 @@
 
 extern crate r0;
 
-use core::fmt;
+use core::{fmt, ptr};
 
 /// Registers stacked (pushed into the stack) during an exception
 #[derive(Clone, Copy)]
@@ -508,6 +508,26 @@ pub unsafe extern "C" fn Reset() -> ! {
 
             trampoline()
         }
+    }
+}
+
+#[doc(hidden)]
+#[no_mangle]
+pub unsafe extern "C" fn DefaultDefaultHandler() {
+    loop {
+        // add some side effect to prevent this from turning into a UDF instruction
+        // see rust-lang/rust#28728
+        ptr::read_volatile(&0u8);
+    }
+}
+
+#[doc(hidden)]
+#[no_mangle]
+pub unsafe extern "C" fn DefaultUserHardFault() {
+    loop {
+        // add some side effect to prevent this from turning into a UDF instruction
+        // see rust-lang/rust#28728
+        ptr::read_volatile(&0u8);
     }
 }
 


korken89:
LGTM

japaric:
@korken89 sorry, I wasn't clear. I meant to get all the team to OK the PR before landing it. I thought adding WIP to the PR title would prevent bors from merging it but it seems I misremembered. We should use a label for this kind of situations -- to block bors from merging the PR.

@therealprof @adamgreig are you OK with the changes? If any of you is unhappy with the implementation we can always revert it or change it.

therealprof:
@japaric I'm happy with the changes.

korken89:
@japaric Ah, sorry about misunderstanding you!  Will be a bit more slow in the future :)

adamgreig:
I'm happy with the changes. It's a shame we have to have the slightly confusing infinite loop. Could we put DefaultDefaultHandler and DefaultUserHardFault in with HardFault in asm.s so we can just write a more-obvious infinite loop without worrying about the optimiser?

> The endless loop is easy to trigger and potentially hard to get out of (even with an attached debugger), e.g. by the SysTick exception and the PendSV. Is that on purpose? Usually I'd expect harmless exceptions and interrupts to be ignored instead.

Isn't that the point of a by-default infinite loop default handler (as per #72)? It hangs, you investigate why and fix it. Would you rather have the DefaultDefaultHandler just return immediately? For users wanting unhandled interrupts to be ignored they can set the DefaultHandler to an empty function (but probably this is a bad idea...).

> OTOH I don't see the UserHardFault handler used anywhere, shouldn't NMI, BusFault, SecureFault, etc. all be using the HardFault handler? Those are ones that should "freeze" the system in the current state for inspection.

NMI, BusFault, SecureFault, etc are different to HardFault in that they're nominally recoverable whereas HardFault is typically not. When those other faults are disabled they default to causing a HardFault. You could still argue that the provided default handler should infinite loop for those faults while returning immediately for something like SysTick (though I'd prefer an infinite loop for all the default handlers).

`UserHardFault` is used by `HardFault` (in `asm.s`) which is in turn placed into the HardFault slot in the vector table by `__EXCEPTIONS`.

therealprof:
> Isn't that the point of a by-default infinite loop default handler (as per #72)? It hangs, you investigate why and fix it.

Why would you want to investigate and fix non-problems? Remember that the default handler is used for **all** interrupts and exceptions and that the handlers and the setup of the NVIC are completely decoupled, meaning it is equally easy to forget enabling an interrupt as it is to forget disabling an interrupt. I usually don't whip out a debugger unless absolutely necessary (i.e. all my typical debugging methods fail) and Rust makes it a breeze to work that way; it's a bit unfortunate if we're going to force this back to the good old fashioned C/C++ way...

My (admittedly silly) understanding was that we're going to loop on faults to preserve the processor state and allow for retrospectively attaching a debugging and potentially clearing the problem on-line (if we issue a `bkpt` and no debugger is attached the system will go into hard fault which can be very tricky to recover from).

> Would you rather have the DefaultDefaultHandler just return immediately?

Yes, as a matter of fact all of my code written so far has done this. My hope was to get rid of this recently introduced boilerplate... ;)

> For users wanting unhandled interrupts to be ignored they can set the DefaultHandler to an empty function (but probably this is a bad idea...).

Why is that?

> NMI, BusFault, SecureFault, etc are different to HardFault in that they're nominally recoverable whereas HardFault is typically not. When those other faults are disabled they default to causing a HardFault. You could still argue that the provided default handler should infinite loop for those faults while returning immediately for something like SysTick (though I'd prefer an infinite loop for all the default handlers).

Yeah, you're of course right, sorry for the bad wording. I should not have mixed up NMI with HardFault and the other faults. The point I was trying to make was: Some faults indicate real problems and you want to halt execution both to let someone know that there **is** a problem and also aid in helping to identify the cause. Regular system exception and interrupts are easy to accidentally set up and trigger but do no harm in any way (and will trigger only once anyway when they're not cleared).

Anyway: I'm not too deeply concerned about this since this reflects good old CMSIS behaviour anyway and can be easily overridden; it's just not very beginner-friendly...

adamgreig:
> Why is that?

I think any unexpected interrupts/exceptions are probably indicative of something you should fix, rather than something you never even found out were happening or were causing surprising bugs elsewhere. If the default behaviour just silently ignored most interrupts and exceptions I think that would lead to more confusion in the end. If you _deliberately_ want to ignore a specific (or all) interrupts that's easy to do, but seems a dangerous default to me.

> My hope was to get rid of this recently introduced boilerplate... ;)

All of mine default to `panic!()` for development, which I guess I'll have to keep as well ;)

> Regular system exception and interrupts are easy to accidentally set up

I don't see this. All the interrupts I've seen require explicit set-up, and all the fault exceptions that are enabled by default are things you really want to know about. If anything I've found the much more common error is intending to set up an interrupt and failing to enable it in both NVIC and the peripheral!

therealprof:
> > Regular system exception and interrupts are easy to accidentally set up

> I don't see this.

I do tinker around a lot with interrupts and also use them for profiling and debugging, e.g. to measure execution times of code with an oscilloscope. It is very easy to forget removing the enabling of an interrupt in the NVIC, a panic or endless loop can easily ruin a few hours of spare time.

> All of mine default to panic!() for development, which I guess I'll have to keep as well ;)

That's reasonable **iff** you always have a debugger attached to each device or write 100% perfect code for 100% perfect hardware running in 100% perfect environments. 😏

> If anything I've found the much more common error is intending to set up an interrupt and failing to enable it in both NVIC and the peripheral!

I certainly agree with that, also forgetting to clear an interrupt can be very annoying. However this doesn't have anything to with those problems and if anything adds more pain points rather than reducing them.

japaric:
> It's a shame we have to have the slightly confusing infinite loop. Could we put DefaultDefaultHandler and DefaultUserHardFault in with HardFault in asm.s so we can just write a more-obvious infinite loop without worrying about the optimiser?

We can but that would be another feature that makes this crate depend on gcc. If we aim to default to rust-lld *and* want people to be able to use quickstart w/o having to install gcc then this crate must build w/o depending on gcc. That would mean putting the UserHardFault feature (access to the stacked registers) behind a Cargo feature and avoiding this suggestion of yours (or put it behind a Cargo feature).

adamgreig:
> We can but that would be another feature that makes this crate depend on gcc.

Right, doesn't seem worth it then.

> the alternative of maintaining the status quo (leaving DefaultHandler and HardFault undefined) and instead having crates like panic_abort to set the default to e.g. panic (exception_panic?)

Given those crates could still exist and override defaults in cortex-m-rt, I think it would be simplest to keep the infinite loop here for minimal boilerplate. If someone wants to make `exception_panic` it would be simple enough, I think? I might do it if no one else does...

Issue #86, opened 2018-08-08
Linker script doesn't put .data into FLASH section, preventing overflow detection


adamgreig:
If you have an application where `.vector_table` + `.text` + `.rodata` fit into `FLASH`, but with `.data` included as well it does not fit, gcc (haven't tried lld) will fail to notice the overflow.

I believe this is because `.data` has an LMA which is a computed address rather than a section:

```c
/* In current link.x.in, note LMA at the top but no use of FLASH section */
  .data : AT(ADDR(.rodata) + SIZEOF(.rodata)) /* LMA */
  {
    *(.data .data.*);

    . = ALIGN(4); /* 4-byte align the end (VMA) of this section */
  } > RAM
```

Instead we might use:
```c
/* Hypothetical alternative example which correctly detects overflow, note "AT > FLASH" */
  .data : ALIGN(4)
  {
      *(.data .data.*)
      . = ALIGN(4);
  } > RAM AT > FLASH
```

I assume there's a reason for doing it the current way (perhaps compatibility with lld? I'm not sure), but anyway failing to detect the overflow should be treated as a bug in the linker script.

The alternative above correctly detects overflow, and seems to do the right thing when everything does fit, but I've not tested it extensively (it's what I'd usually use on a typical embedded C gcc project though). Using `ALIGN(4)` in the declaration of `.data` seems to ensure it's aligned without requiring the `. = ALIGN(4)` we currently have in the end of `.rodata`, too.

japaric:
> perhaps compatibility with lld?

yes, iirc `AT> FLASH` (or was it the `: ALIGN(4)` part?) caused problems with LLD 6.0. New rust-lld is LLD 7.0 so it would be good to check again if this is still a problem. If you send a PR with this change and it passes travis (we now have tests for LLD) I think it would be good to go. Extra points for adding a test that exercises this corner case.

PR #87, opened 2018-08-12
Update linker script to put .data into FLASH

adamgreig:
This is the updated linker script and test from #86.

Sadly it looks like LLD does not use the `ALIGN(4)` in the `.data` section header to align the LMA (unlike gcc), so we still need the `. = ALIGN(4)` at the end of `.rodata` to ensure LMA alignment.

I've added a failing example which the ci script ensures fails when `.data` causes overflow. It just checks for non-zero return from cargo rather than specifically grepping for the linker error messages about `.data` overflowing; since we only have the one failing example I could remove the loop and just specifically check it and grep for the relevant error from gcc and lld if people think that would be better.

### DIFF
--- a/ci/script.sh
+++ b/ci/script.sh
@@ -11,6 +11,9 @@ main() {
         main
         state
     )
+    local fail_examples=(
+        data_overflow
+    )
     if [ $TRAVIS_RUST_VERSION = nightly ]; then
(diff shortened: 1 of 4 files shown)


japaric:
Thanks @adamgreig!

bors r+

Issue #82, opened 2018-08-02
[RFC] [breaking-change] Turn entry!, exception! into attributes


japaric:
What the title says. I believe attributes would be more ergonomic and less redundant. Compare this:

``` rust
entry!(main);

fn main() -> ! {
    // .. user code ..
}

exception!(HardFault, handler);

fn handler() -> ! {
    // .. user code ..
}
```

to this:

``` rust
#[entry]
fn main() -> ! {
    // .. user code ..
}

#[exception(HardFault)]
fn handler() -> ! {
    // .. user code ..
}
```

As a bonus point we get rid of the indirection that the `entry!` macro requires.

`entry!` expands into this for type safety:

``` rust
#[export_name = "main"]
fn __impl_main() -> ! {
    // check the signature of `main`
    let f: fn() -> ! = main;
    f()
}
```

The function call doesn't always get inlined (unless you marked `main` as `#[inline(always)]`, which
looks a bit odd).

On the other hand, `#[entry]` expands into this:

``` rust
#[export_name = "main"]
pub fn main() -> ! {
    // .. user code ..
}
```

so there's no function call even when compiling without optimizations. (The type check is done
syntactically during the macro expansion)

### Unresolved questions

- Choose a syntax for adding state to `#[exception]`s.

---

`#[proc_macro_attribute]`, which is required to *write* these macros, has already been stabilized in
nightly. `#![feature(use_extern_macros)]`, which is required to *use* these macros, has not been
stabilized yet, but it's planned for the 2018 edition (1.31).

japaric:
@rust-embedded/cortex-m this RFC needs a decision.

therealprof:
@japaric I love it but how would you bring it up with the relevant parts still unstable?

adamgreig:
+1 on swapping to the attributes instead of the current macros.

* What do we do for DefaultHandler? `#[exception(*)]` as currently, or a different attribute name?
* Does `svd2rust` swap to using `#[interrupt(...)]`?
* Syntax for the optional state does seem like it might be a bit messy. `#[exception(SysTick, state: u32 = 0)]` is a bit unusual, but I don't see a way around needing to specify the initial value. Having it be an `Option` means you have to check it every interrupt, having it uninitialised is obviously not viable.

japaric:
@therealprof last part of the RFC

> #[proc_macro_attribute], which is required to write these macros, has already been stabilized in
[...]

PR #90, opened 2018-08-18
turn macros into attributes

japaric:
This is a PoC implementation of RFC #82. Assuming that we are OK with this implementation we should
add some compile fail tests (e.g. using a function with the wrong signature as the entry point)
before landing this.

Look at the diff of the examples to get an overview of the changes.

cc @rust-embedded/cortex-m


japaric:
These attributes can be used on the latest nightly without any feature gate (`use_extern_macros` was stabilized yesterday) and you will be able to use them on the next beta.

This is a breaking change because it removes the macros by example.

Instead of directly breaking the rt crate we could have an intermediate test phase where we first publish the rt-macros crate. Then people would be able to test out the attributes by changing their programs in the following way:

``` diff
 #![no_main]
 #![no_std]

-#[macro_use(entry)]
 -extern crate cortex_m_rt as rt;
+extern crate cortex_m_rt_macros as rt;
 extern crate panic_semihosting;

-entry!(main);
+use rt::entry;

+#[entry]
 fn main() -> ! {
     loop {}
 }
```

Though I'm not sure if there's any advantage to doing the release that way.

macros/src/lib.rs
```diff
@@ -0,0 +1,452 @@
...
+                let ty = second.ty;
+                let expr = second.expr;
+                let state = second.ident;
+
+                quote!(
+                    #[export_name = #name_s]
+                    pub fn #ident() {
+                        extern crate cortex_m_rt;
+
+                        cortex_m_rt::Exception::#name;
```
japaric: @adamgreig this is reason why an hypothetical `#[interrupt]` attribute would need the device crate name in its argument. We want to check that the interrupt exists and emit a compiler error if it doesn't and for that we need the full path to the `Interrupt` variant: e.g. `stm32f103xx::Interrupt::EXTI0`.

examples/state.rs
```diff
@@ -5,20 +5,18 @@
...
 fn main() -> ! {
     loop {}
 }
 
 // exception handler with state
-exception!(SysTick, sys_tick, state: u32 = 0);
-
-fn sys_tick(state: &mut u32) {
-    *state += 1;
+#[exception(SysTick, static STATE: u32 = 0)]
```
japaric: An alternative syntax would be to have the second argument be a string. i.e.

``` rust
#[exception(SysTick, "static STATE: u32 = 0")]
fn sys_tick() {}
```

But the current syntax seems to parse without problems and have better spans for error messages. e.g.

``` rust
#[exception(SysTick, static STATE: u32 = true)]
//                                       ~~~~ error: expected `u32`, found `bool`
fn sys_tick() {}
```

therealprof:
@japaric Lovely, I like it!

japaric:
Error messages are terrible by default because all the span information is lost. By this I mean that any compile error caused by the body of e.g. an exception function will be reported with the span of the attribute. Example:

``` rust
#[entry]
// ~~~~~ error: mismatched types, expected `!`, found `()`
fn main() -> ! {
    asm::bkpt()
}
```

To improve you can opt into the `nightly` feature of the `proc-macro2` crate like this:

``` toml
# Cargo.toml
[build-dependencies.proc-macro2]
version = "0.4"
features = ["nightly"]
```

But, as the name of the feature implies, this requires a nightly compiler. The error message then becomes what you would expect:

``` rust
#[entry]
fn main() -> ! {
    asm::bkpt()
//  ~~~~~~~~~~~ error: mismatched types, expected `!`, found `()`
}
```

I don't know if it's planned to stabilize these span improvements for the edition release.

macros/src/lib.rs
adamgreig: Makes sense. I guess we could require that whatever is passed in (to some future `#[interrupt]`) is resolvable to an `Interrupt`, so it could be either a full path, or the user could `use stm32f103::Interrupt::*` and then just call `#[interrupt(EXTI0)]`?

japaric: >  the user could use stm32f103::Interrupt::* and then just call #[interrupt(EXTI0)]?

If we omit the full path that would kind of defeat the purpose of the check. For example, this would compile:

``` rust
struct InterruptThatDoesntExist;

#[interrupt(InterruptThatDoesntExist)]
fn handler() { /* .. */ }
```

Though thinking about it more having *the user* enter the name of the device crate makes it possible to nullify the check. For example,

``` rust
extern crate stm32f103xx; // real device crate

mod stm32f1337 {
    enum Interrupt {
        InterruptThatDoesntExist;
    }
}

#[interrupt(stm32f1337, InterruptThatDoesntExist)]
fn handler() { /* .. */ }
```

japaric:
I have rebased this PR.

@rust-embedded/cortex-m Please vote on RFC #82 by reviewing (approving) this PR.

Amendments to #82 

- The unresolved question about how to support state in the `#[exception]` attribute has been solved with the following syntax:

``` rust
#[exception(SysTick, static STATE: u32 = 0)]
fn sys_tick() { // <- note that the function has no argument
    *STATE += 1; // but that `STATE` can be referenced only within this function
}
```

Also a reminder that both the `proc_macro` and the `use_extern_macro` features has been already stabilized in nightly so you'll be able to use these attributes in the next beta release (2 weeks from now) so I think it's OK to land this right now -- it'll be in a new minor version (v0.6.0) since it's a breaking change.

japaric:
I just thought of an alternative syntax for adding state that may feel less strange:

``` rust
#[exception(SysTick)]
fn sys_tick() {
    static mut STATE: u32 = 0;

    // user code
    *STATE += 1;
}
```

The attribute will magically make `static mut` variables safe (no `unsafe`) to access. It will also cause their type to change to `&mut Type`. The expansion would look like this:

``` rust
#[export_name = "SysTick"]
pub fn sys_tick() {
    // <expansion>
    static mut STATE_: u32 = 0;

    let STATE: &mut u32 = unsafe { &mut STATE_ };
    // </expansion>

    // user code
    *STATE += 1;
}
```

japaric:
And now I'm seeing a problem with the current implementation: it allows data races in safe code. For example:

``` rust
#[exception(SysTick)]
fn sys_tick() {
    static mut STATE: u64 = 0;

    // user code
    *STATE += 1;
}

#[entry]
fn main() -> ! {
    loop {
        sys_tick();
    }
}
```

Here `main` and SysTick handler race for access to `STATE` -- this is UB.

The fix is to either (a) make `sys_tick` unsafe to call, or (b) make `sys_tick` impossible to call. If we do the latter then we can simplify the attribute a bit:

``` rust
#[exception]
fn SysTick() {
    static mut STATE: u32 = 0;

    // user code
    *STATE += 1;
}

#[entry]
fn main() -> ! {
    loop {
        SysTick(); //~ error: this function doesn't exist
    }
}
```

We can use the function name as the exception name; that way we don't need an argument. The above code would expand to:

``` rust
#[export_name = "SysTick"]
pub fn cf1dglqbt3jr7uiu() { // some randomly generated string that the user can't rely on
    // <expansion>
    static mut STATE_: u32 = 0;

    let STATE: &mut u32 = unsafe { &mut STATE_ };
    // </expansion>

    // user code
    *STATE += 1;
}

#[entry]
fn main() -> ! {
    SysTick(); // error: this function doesn't exist
}
```

korken89:
I'm not sure I'm a fan of "magically" making statics safe, but I also do not have any good input for a better solution, though the original idea was quite nice on the eyes and properly conveys that there is a state.
Why not keep this syntax? I can't quite see the issue with it.

japaric:
@korken89 with the `static` in the attribute it's not clear that the `static` is only visible to the function or if it's in the global / outer scope since it's between those two spaces. Also there's no precedent for attributes that have `static`s as arguments so that syntax will look odd to most people.

With the static inside the function it's immediately clear that it can't be referred to from outside the function. I believe that making it safe to access it's not surprising exactly because it's local to the function.

korken89:
I see, and agree. Having the static in the function would indeed be cleaner.

therealprof:
@japaric Yes, please! The local state definition for the interrupt handlers was very weird in the past and caused lots of weird syntactic problems with non-integral types which should be all gone with the new syntax. This is very similar to something a C/C++ programmer might do, too, so that's also a big plus. And last but not least this reduces the boilerplating quite a bit and makes things safer (by preventing calling an interrupt handler from another function) so goodies all around!

japaric:
OK, I have made the exception handlers uncallable and also change the syntax of `#[exception]`: no arguments required and the state must now be declared as `static mut` variables at the beginning of the function body.

japaric:
Also, we can use the `static mut` trick in `#[entry]` too. That will turn local `static mut` variables into `&'static mut` references. We would have to make the `#[entry]` function uncallable as well.

japaric:
latest commit makes it safe to access local `static mut` variables in the entry point.

@therealprof @adamgreig @korken89 in #82 it sounded like you were in favor of this change. Could you confirm?

@thejpster @ithinuel I'd love to get your input on RFC #82 and this implementation. Are you in favor of that RFC or do you have concerns about it?

adamgreig:
I really like the new no-argument attributes and the new syntax for state. Having mutable statics work in `entry` would solve a bunch of nuisance problems I've had in the past with initialising statics, for example setting up buffers for smoltcp. As far as I can tell this will actually be extremely useful for a lot of embedded development and maybe warrants a little more exposition in the documentation, but that can come later when people have a bit more experience with it.

:+1: from me!

japaric:
@ithinuel 

The proc macros I'm using here are planned to be stable by the 2018 edition; they don't require feature gates in nightly; and they won't require features gates in RC1 (next beta) and RC2 (next next beta).

---

Alright, thanks for the input everyone. We have majority so this is officially accepted.

I'm going to add some compile fail tests then any of you can review and land this.

japaric:
This should be ready to land, but while adding the compile-fail tests I noticed that we could relax some constraints. I'll leave some inline comments about that.

tests/compile-fail/default-handler-bad-signature-2.rs
```diff
@@ -0,0 +1,16 @@
...
+use cortex_m_rt::{entry, exception};
+
+#[entry]
+fn foo() -> ! {
+    loop {}
+}
+
+#[exception] //~ ERROR custom attribute panicked
+//~^ HELP `DefaultHandler` exception must have signature `fn(i16)`
+unsafe fn DefaultHandler(_irqn: i16) {}
```
japaric: Do we want to allow this? I don't see a problem with letting the user state this is `unsafe` because this function can't be called from software anyways.

tests/compile-fail/default-handler-bad-signature-3.rs
```diff
@@ -0,0 +1,18 @@
...
+use cortex_m_rt::{entry, exception};
+
+#[entry]
+fn foo() -> ! {
+    loop {}
+}
+
+#[exception] //~ ERROR custom attribute panicked
+//~^ HELP `DefaultHandler` exception must have signature `fn(i16)`
+fn DefaultHandler(_irqn: i16) -> ! {
```
japaric: Do we want to allow this? I see no problem with letting this be a divergent function.

tests/compile-fail/entry-bad-signature-3.rs
```diff
@@ -0,0 +1,13 @@
...
+#![no_std]
+
+extern crate cortex_m_rt;
+extern crate panic_semihosting;
+
+use cortex_m_rt::entry;
+
+#[entry] //~ ERROR custom attribute panicked
+//~^ HELP `#[entry]` function must have signature `fn() -> !`
+unsafe fn foo() -> ! {
```
japaric: Do we want to allow this? Same rationale as allowing `unsafe` `DefaultHandler`.

tests/compile-fail/exception-bad-signature-2.rs
```diff
@@ -0,0 +1,16 @@
...
+use cortex_m_rt::{entry, exception};
+
+#[entry]
+fn foo() -> ! {
+    loop {}
+}
+
+#[exception] //~ ERROR custom attribute panicked
+//~^ HELP `#[exception]` functions other than `DefaultHandler` and `HardFault` must have signature `fn()`
+unsafe fn SysTick() {}
```
japaric: Do we want to allow this? Same rationale as `unsafe` `DefaultHandler`

tests/compile-fail/exception-bad-signature-3.rs
```diff
@@ -0,0 +1,18 @@
...
+use cortex_m_rt::{entry, exception};
+
+#[entry]
+fn foo() -> ! {
+    loop {}
+}
+
+#[exception] //~ ERROR custom attribute panicked
+//~^ HELP `#[exception]` functions other than `DefaultHandler` and `HardFault` must have signature `fn()`
+fn SysTick() -> ! {
```
japaric: Do we want to allow this? Same rationale as divergent `DefaultHandler`

tests/compile-fail/hard-fault-bad-signature-2.rs
```diff
@@ -0,0 +1,18 @@
...
+use cortex_m_rt::{entry, exception, ExceptionFrame};
+
+#[entry]
+fn foo() -> ! {
+    loop {}
+}
+
+#[exception] //~ ERROR custom attribute panicked
+//~^ HELP `HardFault` exception must have signature `fn(&ExceptionFrame) -> !`
+unsafe fn HardFault(_ef: &ExceptionFrame) -> ! {
```
japaric: Do we want to allow this? Same rationale as `unsafe` `DefaultHandler`

tests/compile-fail/default-handler-bad-signature-2.rs
adamgreig: :+1: from me, no reason to not permit this, and I wouldn't be surprised if a lot of unsafe code in embedded projects is in the ISRs.

adamgreig:
Rather than commenting inline on each one: I'm happy to allow all of those unsafe/divergent functions. There's no safety reasons not to, so I'd rather let the user decide.

japaric:
Ok, I have relaxed the signature checks and added more compile-fail tests and examples (which I'm using as compile-pass tests). I have also made the reachability requirement more prominent in the docs.

adamgreig:
Looks great! I've gone through the new code and also tested this branch on a few small examples on hardware with exceptions and pre_init and everything seems to work as expected.

bors r+

PR #95, opened 2018-08-25
[RFC] remove build dependency on arm-none-eabi-gcc (binary blob alternative)

### LINKED ISSUE #91: [RFC] [breaking-change] don't depend on GCC by default
It's very likely that the thumb targets will soon switch to rust-lld as their default linker (see
rust-embedded/wg#160). If we want to achieve the goal of *not* depending on GCC, by default, to
build Cortex-M applications then we also have to make core crates like this one not depend on GCC by
default. This RFC is exactly about that.

I propose that in the next minor (breaking) release, v0.6.0, we add an opt-in Cargo feature named
"exception-frame" that changes the signature of the HardFault handler as it's shown below:

``` rust
// "exception-frame" disabled
#[exception(HardFault)]
fn handler(/* no `ExceptionFrame` here */) -> ! {
    // ..
}

// "exception-frame" enabled
#[exception(HardFault)]
fn handler(ef: &ExceptionFrame) -> ! {
    // ..
}
```

The rationale for this change is that providing `ExceptionFrame` information to the handler requires
[an external assembly file](https://github.com/rust-embedded/cortex-m-rt/blob/v0.5.2/asm.s) and that
makes this crate depend on GCC. By making this information opt-in this crate would, by default, work
w/o GCC being present. If the user wants the `ExceptionFrame` information they'll need to have GCC
installed.

cc @rust-embedded/cortex-m @hannobraun @crawford

japaric:
Before this commit we used gcc to assemble external assembly files into object
files that we linked into our Rust program.

This commit drops the dependency on gcc by shipping the already assembled object
files with this crate source code.

---

This is an alternative to RFC #91 that doesn't require a breaking change or
adding a new Cargo feature and can be implemented right now.

See #91 for the rationale of dropping the dependency on gcc.

This approach can be applied to other Cortex-M crates like cortex-m-semihosting
and cortex-m (would subsume RFC rust-embedded/cortex-m#107).

This seems like an overall better approach to me, but before I go opening more
PRs I want to hear your thoughts, @rust-embedded/cortex-m

closes #91


therealprof:
Very nice, I like it!

korken89:
I agree, this looks good!

adamgreig:
> IMO, the best long term solution is to have asm! and global_asm! in stable Rust.

:+1:

> I'm not too keen on adding a public Cargo feature that's only useful while developing (and not in all the cases).

That's fair. I'm still a little uneasy shipping binary images as part of a foundational crate with no easy way to say "please build these from the source actually". But it's not a show-stopper and this way is better than the alternatives.

Is it possible to ensure the built binaries are up-to-date before `cargo publish` bundles them up? Are the builds reproducible / could we have Travis rebuild the `.a` files and check they match whatever's in Git, to prevent updated `.s` files being committed (or worse, published) alongside outdated `.a` files?

japaric:
> Is it possible to ensure the built binaries are up-to-date before cargo publish bundles them up?

@adamgreig done. See https://travis-ci.org/rust-embedded/cortex-m-rt/jobs/420519538#L541

thejpster:
Putting binaries in the source tree goes against all my principles, but actually, it may be the least bad option here. Provided we have those shell scripts for rebuilding, I don't see the need to bake it into the Cargo.toml because it's not something people will do often.

therealprof:
@thejpster I agree that binaries are awkward, however since we ship the source  and it's trivial to build yourself it's not too bad in this particular case.

PR #99, opened 2018-08-28
Added mdcheckr to detect broken links in Markdown 

korken89:
This will be the PR for checking for broken links in documentation, I will start adding it to more places as soon as we are happy with it here.
Please come with comments if there is something more that you'd like to add or change.

### DIFF
--- a/ci/script.sh
+++ b/ci/script.sh
@@ -78,7 +78,7 @@ main() {
         ./check-blobs.sh
 
         # Check Markdown files for broken links
-        find . -name '*.md' -print0 | xargs -0 mdcheckr
+        mdcheckr **/*.md
     fi
 }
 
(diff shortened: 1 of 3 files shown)


ci/script.sh
```diff
@@ -76,6 +76,9 @@ main() {
 
     if [ $TARGET = x86_64-unknown-linux-gnu ]; then
         ./check-blobs.sh
+
+        # Check Markdown files for broken links
+        find . -name '*.md' -print0 | xargs -0 mdcheckr
```
japaric: is this the same as `mdcheckr **/*.md`? If yes, I find the glob version more readable.

korken89:
Any comments on the CHANGELOG link issue @rust-embedded/cortex-m ? I am unable to get around the scenario @japaric mentioned, unless removing the file from being checked. But it sub-optimal.

korken89:
So eh, should we close this?

therealprof:
@korken89 Sounds good to me.

korken89:
If we want to look into this again we can redo it

PR #101, opened 2018-08-31
[do not merge] Fixing the Travis build

korken89:
The travis build is failing on nightly, probably the old cortex-m dependency that's the issue.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ r0 = "0.2.1"
 [dev-dependencies]
 panic-semihosting = "0.3.0"
 panic-abort = "0.2.0"
-cortex-m = "0.5.4"
+cortex-m = "0.5.6"
 
 [features]
 device = []


korken89:
1. It seems to be that the overflowing examples fail (as they should) but the `!` does not make the fail a pass in the script. Does anyone know why this is happening?
2. The final error is referring to `rust-lld: error while loading shared libraries: libLLVM-7.so: cannot open shared object file: No such file or directory`, but I am unable to replicate this on my local machine. Any ideas?

cc @rust-embedded/cortex-m

japaric:
(2) is due to rust-lang/rust#53813. It will be fixed in tomorrow's nightly. The workaround is to install the `llvm-tools-preview` component or to roll back to a 2-days old nightly.

korken89:
Thanks for the clarification!

japaric:
Restarted the CI build and [it passed](https://travis-ci.org/rust-embedded/cortex-m-rt/builds/423038406?utm_source=github_status&utm_medium=notification) so I'm going to close this.

korken89:
Thanks @japaric !

PR #104, opened 2018-09-06
v0.6.0, take 2

japaric:
the dev-dependency of cortex-m-rt-macros needs to specify a version or can't be
published; this commit fixes that

docs.rs won't be able to build docs for these two crates so this also changes
the documentation link to https://rust-embedded.github.io/cortex-m-rt/ where I
have (manually) implemented a poor man's docs.rs

now we should be able to publish a new version

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,11 +2,11 @@
 authors = ["Jorge Aparicio <jorge@japaric.io>"]
 categories = ["embedded", "no-std"]
 description = "Minimal runtime / startup for Cortex-M microcontrollers"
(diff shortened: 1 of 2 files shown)


japaric:
This is awesome. We can't publish cortex-m-rt v0.6.0 because it depends on cortex-m-rt-macros v0.1.0, but we can't publish cortex-m-rt-macros v0.1.0 either because it depends on cortex-m-rt v0.6.0. :joy: 

I'll cheat a bit.

japaric:
https://crates.io/crates/cortex-m-rt/0.6.0 :tada:

PR #106, opened 2018-09-06
error during compilation if two copies of cortex-m-rt are being linked

japaric:
linking two copies into the final binary produces a confusing linker error
message. This improves the situation by producing an error at compile time.

This will have to be backported into the v0.5.x series or you won't get the new
error message.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ license = "MIT OR Apache-2.0"
 name = "cortex-m-rt"
 readme = "README.md"
 repository = "https://github.com/japaric/cortex-m-rt"
-version = "0.6.0"
+version = "0.6.1"
 
 [dependencies]
 r0 = "0.2.1"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -8,7 +8,7 @@ license = "MIT OR Apache-2.0"
 name = "cortex-m-rt"
 readme = "README.md"
 repository = "https://github.com/japaric/cortex-m-rt"
```
therealprof: This should probably be changed to https://github.com/rust-embedded/cortex-m-rt at some point?

therealprof:
I've a feeling we're going to bump the version at least one more time today. Maybe be careful with single change version bumps? ;)

japaric:
Published:

- https://crates.io/crates/cortex-m-rt/0.6.1
- https://crates.io/crates/cortex-m-rt/0.5.4

japaric:
Thinking more about it: v0.6.x still uses the same symbols and linker sections as v0.5.x so instead of making it an error to use both we could use the [semver-trick](https://github.com/dtolnay/semver-trick) to make them work together (i.e. having both minor versions in the dependency graph doesn't cause any error). I'll try that and see if it works ...

japaric:
Published: https://crates.io/crates/cortex-m-rt/0.5.7 which is compatible with v0.6.x

PR #107, opened 2018-09-06
Disable default-features on rand dependency to avoid std version

### LINKED ISSUE #105: New cortex-m-rt rand dependency breaks builds
I was just trying to port the `microbit` crate to `cortex-m-rt` 0.6.0 but I'm facing this here:
```
   Compiling rand_core v0.2.1
   Compiling cortex-m-rt v0.6.0
error[E0463]: can't find crate for `std`
  |
  = note: the `thumbv6m-none-eabi` target may not be installed

For more information about this error, try `rustc --explain E0463`.
error: Could not compile `rand_core`.
warning: build failed, waiting for other jobs to finish...
error: build failed
```

@japaric Seems something is messing (or not) with the required flags for `no_std`.

therealprof:
Due to the single dependency tree, the attempted use of a std version
flips all depending crates to the std version as well which will not
compile on no_std systems.

Fixes #105

### DIFF
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -4,12 +4,18 @@ extern crate proc_macro;
 extern crate rand;
 #[macro_use]
 extern crate quote;
+extern crate core;
 extern crate proc_macro2;
 extern crate syn;
 
 use proc_macro2::Span;
 use rand::Rng;
+use rand::SeedableRng;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::time::{SystemTime, UNIX_EPOCH};
 use syn::{FnArg, Ident, Item, ItemFn, ItemStatic, ReturnType, Stmt, Type, Visibility};
+
+static CALL_COUNT: AtomicUsize = AtomicUsize::new(0);
 
 use proc_macro::TokenStream;
 
@@ -492,7 +498,23 @@ pub fn pre_init(args: TokenStream, input: TokenStream) -> TokenStream {
 
 // Creates a random identifier
 fn random_ident() -> Ident {
-    let mut rng = rand::thread_rng();
+    let secs = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .unwrap()
+        .as_secs();
+
+    let count: u64 = CALL_COUNT.fetch_add(1, Ordering::SeqCst) as u64;
+    let mut seed: [u8; 16] = [0; 16];
+
+    for (i, v) in seed.iter_mut().take(8).enumerate() {
+        *v = ((secs >> (i * 8)) & 0xFF) as u8
+    }
+
+    for (i, v) in seed.iter_mut().skip(8).enumerate() {
+        *v = ((count >> (i * 8)) & 0xFF) as u8
+    }
+
+    let mut rng = rand::rngs::SmallRng::from_seed(seed);
     Ident::new(
         &(0..16)
             .map(|i| {
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,10 @@ cortex-m = "0.5.4"
 panic-abort = "0.3.0"
 panic-semihosting = "0.4.0"
 
+[dev-dependencies.rand]
+default-features = false
+version = "0.5.5"
+
 [target.'cfg(not(target_os = "none"))'.dev-dependencies]
 compiletest_rs = "0.3.14"
 
--- a/ci/script.sh
+++ b/ci/script.sh
@@ -20,11 +20,12 @@ main() {
         minimal
         override-exception
         pre_init
+        rand
         state
         unsafe-default-handler
-        unsafe-hard-fault
         unsafe-entry
         unsafe-exception
+        unsafe-hard-fault
     )
     local fail_examples=(
         data_overflow
--- a/examples/rand.rs
+++ b/examples/rand.rs
@@ -1,0 +1,24 @@
+//! Use rand crate to ensure it's configured for no_std compatbility
+
+#![deny(warnings)]
+#![no_main]
+#![no_std]
+
+extern crate cortex_m_rt as rt;
+use rt::entry;
+
+extern crate panic_semihosting;
+
+extern crate rand;
+use rand::Rng;
+use rand::SeedableRng;
+
+// the program entry point
+#[entry]
+fn main() -> ! {
+    let seed: [u8; 32] = [0; 32];
+    let mut rng = rand::ChaChaRng::from_seed(seed);
+    let _ = rng.gen::<u32>();
+
+    loop {}
+}
--- a/macros/Cargo.toml
+++ b/macros/Cargo.toml
@@ -14,12 +14,15 @@ proc-macro = true
 
 [dependencies]
 quote = "0.6.6"
-rand = "0.5.5"
 proc-macro2 = "0.4.15"
 
 [dependencies.syn]
 features = ["extra-traits", "full"]
 version = "0.14.8"
+
+[dependencies.rand]
+version = "0.5.5"
+default-features = false
 
 [dev-dependencies]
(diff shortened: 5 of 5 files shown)


japaric:
LGTM. Could you add an example that links to rand with default features disabled? That should serve as a regression test.

japaric:
note that to get the example to build on CI you'll have to add it: https://github.com/rust-embedded/cortex-m-rt/blob/5baa082ff147a3b90fa00d09de488490be8b5e93/ci/script.sh#L14-L28

therealprof:
Shouldn't any example do? The rand stuff is used for generation of unique linker symbols, no?

therealprof:
@japaric Please have a look now.

japaric:
Hmm, the build is failing since the first commit with: "error[E0658]: use of unstable library feature 'rustc_private'", and I don't understand why ...

therealprof:
😀. And now travis-ci is running into the exact same problem I mentioned in the ticket before...

therealprof:
I think it's because the version seems to be newer than the upstream version so it uses that instead which triggers this problem. As I said: I'm having the same troubles locally.

therealprof:
I simply compiled the example via `cargo build --example rand`. Worked without problems.

therealprof:
LOL and of course once I fixed the type it does not compile anymore. 😀

Raining lemons today...

japaric:
@therealprof you didn't add `#![feature(rustc_private)]` to `macros/src/lib.rs` right? That would "fix" the problem by making the crate using the `rand` shipped with the toolchain but we can't use that because it's unstable.

therealprof:
That's indeed what I did. Do you want me to rewrite the rand stuff in macros to use the seedable rng which works in no_std?

japaric:
> Do you want me to rewrite the rand stuff in macros to use the seedable rng which works in no_std?

Yes, please. Though, I'm concerned that using a seed which is not randomly generated will result in (a) predictable identifiers which then the user can the use to call exception handlers breaking the non-reentrant invariant, and / or (b) collisions between `#[entry] fn foo() -> ! { .. }` and `#[pre_init] unsafe fn foo() {}` if the seed is derived from the identifier.

therealprof:
@japaric Well, the stuff is not really cryptographic anyway. If someone wants to desperately do bad stuff with the entry points, they can still do it. Collisions would be indeed a problem but a `rand::SeedableRng` is not going to have more of a collision than the threaded rng. We could use `mem::uninitialized()` to see the RNG, that would make it slightly more unpredictable, if you're worried about reusing the same seed over and over again.

japaric:
> If someone wants to desperately do bad stuff with the entry points, they can still do it.

They can't with the current implementation.

If they call `cargo expand` and see that `entry` was renamed to `bsb49kvt39ff0w29` and then modify their code to call `bsb49kvt39ff0w29` it won't compile next time because the attribute will generate a *different* identifier. The key property here is that the attribute produces a different identifier on each run. It's not important what RNG is used; it's important that the *seed* of the RNG is different on *each run*.

> We could use mem::uninitialized() to seed the RNG

That could turn into UB in any future LLVM release so it's not an option.

adamgreig:
Can we just prevent the function from being called from safe code? I get that we can't just hide it in a private module (sadly, as that seems like it would totally fix the issue) but perhaps we can stick it inside an unsafe function wrapper with a predictable name? Or any other options to avoid rand?

Otherwise... seed with the build time... :smiling_imp:

therealprof:
> If they call cargo expand and see that entry was renamed to bsb49kvt39ff0w29 and then modify their code to call bsb49kvt39ff0w29 it won't compile next time because the attribute will generate a different identifier. 

Sure... but then again if someone *really* wants to do it they can do still it after the fact, i.e. without recompiling from source.

> it's important that the seed of the RNG is different on each run.

Well, it's a new feature anyway. 😏

> That could turn into UB in any future LLVM release so it's not an option.

Uhm....

> Otherwise... seed with the build time... 😈

Or address of something in the running process. ;)

japaric:
> perhaps we can stick it inside an unsafe function wrapper with a predictable name?

That's an option but we would need a trampoline to preserve the safety that the user declared:

```
#[exception]
fn SysTick() { .. }

// expands into
#[no_mangle]
pub unsafe fn SysTic() {
    // copy paste the user function here
    fn SysTick() { .. }

    SysTick()
}
```

> Otherwise... seed with the build time... 

Seed with `Instant::now().nanos()` or something like that, maybe.

japaric:
> Uhm....

You should read through the Unsafe Code Guidelines team discussions. Apparently most `unsafe` operations are UB now. In any case, `mem::uninitialized` is going to be deprecated in favor of `MaybeUninit` and it's UB to read out its inner value if it has not been previously initialized.

therealprof:
... and of course the Rng is seeded on each invocation of the random token function while the seed is only increasing each second which one fast machines means: crap (and I've been able to experience it first hand). I've added an atomic counter, increased on each invocation (much easier than the pro way of initialising once and reusing the same Rng), to continuously change the seed. Now collisions are a thing of the past.

macros/src/lib.rs
```diff
@@ -6,10 +6,17 @@ extern crate rand;
 extern crate quote;
 extern crate proc_macro2;
 extern crate syn;
+extern crate core;
 
+use core::mem;
```
japaric: Why `core::mem` and not `std::mem`?

macros/src/lib.rs
```diff
@@ -492,7 +499,10 @@ pub fn pre_init(args: TokenStream, input: TokenStream) -> TokenStream {
 
 // Creates a random identifier
 fn random_ident() -> Ident {
-    let mut rng = rand::thread_rng();
+    let seed = SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_secs();
+    CALL_COUNT.fetch_add(1, Ordering::SeqCst);
```
japaric: This returns the previous value. Why not use that instead of doing another atomic load?

macros/src/lib.rs
```diff
@@ -492,7 +499,10 @@ pub fn pre_init(args: TokenStream, input: TokenStream) -> TokenStream {
...
+    let seed = [seed, CALL_COUNT.load(Ordering::SeqCst) as u64];
+    let mut rng = rand::rngs::SmallRng::from_seed( unsafe { mem::transmute(seed) } );
```
japaric: Let's not use `unsafe` code here. There are methods to go from `u64` to `[u8; 8]` and vice versa that could be used here.

adamgreig:
Could we use `.subsec_nanos()` instead of `.as_secs()` to avoid the issue? I'm not sure which systems have `std` but don't give nanosecond-level timers...

macros/src/lib.rs
therealprof: To the best of my knowledge only in nightly...
https://doc.rust-lang.org/nightly/std/primitive.u64.html#method.to_be_bytes

Of course I could do some bitmasking. ;)

therealprof:
> Could we use .subsec_nanos() instead of .as_secs() to avoid the issue? I'm not sure which systems have std but don't give nanosecond-level timers...

Where did you dig that out? 😲
However that only yields a 32 bit value so that doesn't simplify things...

macros/src/lib.rs
japaric: > To the best of my knowledge only in nightly...

Still? I thought it was already stabilized. Perhaps I was thinking of `f32.to_bits`.

> Of course I could do some bitmasking. ;)

I would prefer that over unsafe code.

adamgreig:
I was more thinking you could seed it once with that and then not worry about the counter. Though I guess you have to do the same twiddling to turn it into `[u8; 8]`.

therealprof:
So... bit twiddling it is.

korken89:
Great work!

PR #111, opened 2018-09-07
reduce the .text size of RAM initialization by ~500 bytes

### LINKED ISSUE #110: RAM initialization routines are huge
when compiling with optimizations because LLVM loop unrolls zero_bss and init_data. This causes binaries to be about 500 bytes bigger in size compared to non-loop unrolled version. Here are the disassemblies:

NOTE: I have inserted BKPTs before and after zero_bss and init_data

- `opt-level = "z"`

``` console
$ size -A app
.text                 78       0x400
```

``` armasm
00000400 <Reset>:
 400:   f000 f820       bl      444 <DefaultPreInit>
 404:   4809            ldr     r0, [pc, #36]   ; (42c <_stext+0x2c>)
 406:   2200            movs    r2, #0
 408:   4909            ldr     r1, [pc, #36]   ; (430 <_stext+0x30>)
 40a:   be00            bkpt    0x0000
 40c:   e000            b.n     410 <Reset+0x10>
 40e:   c004            stmia   r0!, {r2}
 410:   4288            cmp     r0, r1
 412:   d3fc            bcc.n   40e <Reset+0xe>
 414:   4807            ldr     r0, [pc, #28]   ; (434 <_stext+0x34>)
 416:   be00            bkpt    0x0000
 418:   4907            ldr     r1, [pc, #28]   ; (438 <_stext+0x38>)
 41a:   4a08            ldr     r2, [pc, #32]   ; (43c <_stext+0x3c>)
 41c:   e001            b.n     422 <Reset+0x22>
 41e:   c908            ldmia   r1!, {r3}
 420:   c008            stmia   r0!, {r3}
 422:   4290            cmp     r0, r2
 424:   d3fb            bcc.n   41e <Reset+0x1e>
 426:   be00            bkpt    0x0000
 428:   e7fe            b.n     428 <Reset+0x28>
 42a:   bf00            nop
 42c:   20000000        .word   0x20000000
 430:   20000000        .word   0x20000000
 434:   20000000        .word   0x20000000
 438:   00000450        .word   0x00000450
 43c:   20000000        .word   0x20000000
```

- `opt-level = "3"` (default)

``` console
$ size -A app
.text                594       0x400
```

``` armasm
00000400 <Reset>:
 400:	f000 f922 	bl	648 <DefaultPreInit>
 404:	f240 0000 	movw	r0, #0
 408:	f240 0200 	movw	r2, #0
 40c:	f2c2 0000 	movt	r0, #8192	; 0x2000
 410:	f2c2 0200 	movt	r2, #8192	; 0x2000
 414:	4282      	cmp	r2, r0
 416:	be00      	bkpt	0x0000
 418:	d25a      	bcs.n	4d0 <Reset+0xd0>
 41a:	1d11      	adds	r1, r2, #4
 41c:	ea6f 0c02 	mvn.w	ip, r2
 420:	4281      	cmp	r1, r0
 422:	f04f 0401 	mov.w	r4, #1
 426:	460b      	mov	r3, r1
 428:	bf38      	it	cc
 42a:	4603      	movcc	r3, r0
 42c:	4463      	add	r3, ip
 42e:	eb04 0393 	add.w	r3, r4, r3, lsr #2
 432:	079b      	lsls	r3, r3, #30
 434:	4613      	mov	r3, r2
 436:	d024      	beq.n	482 <Reset+0x82>
 438:	f04f 0e00 	mov.w	lr, #0
 43c:	4613      	mov	r3, r2
 43e:	f843 eb04 	str.w	lr, [r3], #4
 442:	4283      	cmp	r3, r0
 444:	461d      	mov	r5, r3
 446:	bf38      	it	cc
 448:	4605      	movcc	r5, r0
 44a:	4465      	add	r5, ip
 44c:	eb04 0495 	add.w	r4, r4, r5, lsr #2
 450:	f004 0403 	and.w	r4, r4, #3
 454:	2c01      	cmp	r4, #1
 456:	d014      	beq.n	482 <Reset+0x82>
 458:	4613      	mov	r3, r2
 45a:	2701      	movs	r7, #1
 45c:	f843 ef04 	str.w	lr, [r3, #4]!
 460:	4283      	cmp	r3, r0
 462:	bf38      	it	cc
 464:	4603      	movcc	r3, r0
 466:	4463      	add	r3, ip
 468:	eb07 0393 	add.w	r3, r7, r3, lsr #2
 46c:	f003 0303 	and.w	r3, r3, #3
 470:	2b02      	cmp	r3, #2
 472:	d004      	beq.n	47e <Reset+0x7e>
 474:	2300      	movs	r3, #0
 476:	6093      	str	r3, [r2, #8]
 478:	f102 030c 	add.w	r3, r2, #12
 47c:	e001      	b.n	482 <Reset+0x82>
 47e:	f102 0308 	add.w	r3, r2, #8
 482:	4281      	cmp	r1, r0
 484:	bf38      	it	cc
 486:	4601      	movcc	r1, r0
 488:	4461      	add	r1, ip
 48a:	0889      	lsrs	r1, r1, #2
 48c:	2903      	cmp	r1, #3
 48e:	d31f      	bcc.n	4d0 <Reset+0xd0>
 490:	2100      	movs	r1, #0
 492:	f103 0210 	add.w	r2, r3, #16
 496:	6019      	str	r1, [r3, #0]
 498:	4282      	cmp	r2, r0
 49a:	6059      	str	r1, [r3, #4]
 49c:	6099      	str	r1, [r3, #8]
 49e:	60d9      	str	r1, [r3, #12]
 4a0:	bf3f      	itttt	cc
 4a2:	6011      	strcc	r1, [r2, #0]
 4a4:	6051      	strcc	r1, [r2, #4]
 4a6:	6091      	strcc	r1, [r2, #8]
 4a8:	60d1      	strcc	r1, [r2, #12]
 4aa:	bf3c      	itt	cc
 4ac:	3210      	addcc	r2, #16
 4ae:	4282      	cmpcc	r2, r0
 4b0:	d20e      	bcs.n	4d0 <Reset+0xd0>
 4b2:	6011      	str	r1, [r2, #0]
 4b4:	6051      	str	r1, [r2, #4]
 4b6:	6091      	str	r1, [r2, #8]
 4b8:	60d1      	str	r1, [r2, #12]
 4ba:	3210      	adds	r2, #16
 4bc:	4282      	cmp	r2, r0
 4be:	d207      	bcs.n	4d0 <Reset+0xd0>
 4c0:	f102 0310 	add.w	r3, r2, #16
 4c4:	6011      	str	r1, [r2, #0]
 4c6:	4283      	cmp	r3, r0
 4c8:	6051      	str	r1, [r2, #4]
 4ca:	6091      	str	r1, [r2, #8]
 4cc:	60d1      	str	r1, [r2, #12]
 4ce:	d3e0      	bcc.n	492 <Reset+0x92>
 4d0:	f240 0e00 	movw	lr, #0
 4d4:	f240 0000 	movw	r0, #0
 4d8:	f2c2 0e00 	movt	lr, #8192	; 0x2000
 4dc:	f2c2 0000 	movt	r0, #8192	; 0x2000
 4e0:	4570      	cmp	r0, lr
 4e2:	be00      	bkpt	0x0000
 4e4:	f080 8089 	bcs.w	5fa <Reset+0x1fa>
 4e8:	1d02      	adds	r2, r0, #4
 4ea:	ea6f 0c00 	mvn.w	ip, r0
 4ee:	4572      	cmp	r2, lr
 4f0:	f04f 0501 	mov.w	r5, #1
 4f4:	4611      	mov	r1, r2
 4f6:	f240 6454 	movw	r4, #1620	; 0x654
 4fa:	bf38      	it	cc
 4fc:	4671      	movcc	r1, lr
 4fe:	f2c0 0400 	movt	r4, #0
 502:	4461      	add	r1, ip
 504:	eb05 0191 	add.w	r1, r5, r1, lsr #2
 508:	0789      	lsls	r1, r1, #30
 50a:	d027      	beq.n	55c <Reset+0x15c>
 50c:	4621      	mov	r1, r4
 50e:	4603      	mov	r3, r0
 510:	f851 6b04 	ldr.w	r6, [r1], #4
 514:	f843 6b04 	str.w	r6, [r3], #4
 518:	4573      	cmp	r3, lr
 51a:	461e      	mov	r6, r3
 51c:	bf38      	it	cc
 51e:	4676      	movcc	r6, lr
 520:	4466      	add	r6, ip
 522:	eb05 0596 	add.w	r5, r5, r6, lsr #2
 526:	f005 0503 	and.w	r5, r5, #3
 52a:	2d01      	cmp	r5, #1
 52c:	d018      	beq.n	560 <Reset+0x160>
 52e:	6861      	ldr	r1, [r4, #4]
 530:	4603      	mov	r3, r0
 532:	f843 1f04 	str.w	r1, [r3, #4]!
 536:	4573      	cmp	r3, lr
 538:	bf38      	it	cc
 53a:	4673      	movcc	r3, lr
 53c:	eb03 010c 	add.w	r1, r3, ip
 540:	2301      	movs	r3, #1
 542:	eb03 0191 	add.w	r1, r3, r1, lsr #2
 546:	f001 0103 	and.w	r1, r1, #3
 54a:	2902      	cmp	r1, #2
 54c:	d067      	beq.n	61e <Reset+0x21e>
 54e:	68a1      	ldr	r1, [r4, #8]
 550:	f100 030c 	add.w	r3, r0, #12
 554:	6081      	str	r1, [r0, #8]
 556:	f104 010c 	add.w	r1, r4, #12
 55a:	e001      	b.n	560 <Reset+0x160>
 55c:	4603      	mov	r3, r0
 55e:	4621      	mov	r1, r4
 560:	4572      	cmp	r2, lr
 562:	bf38      	it	cc
 564:	4672      	movcc	r2, lr
 566:	eb02 000c 	add.w	r0, r2, ip
 56a:	0880      	lsrs	r0, r0, #2
 56c:	2803      	cmp	r0, #3
 56e:	d344      	bcc.n	5fa <Reset+0x1fa>
 570:	f103 0210 	add.w	r2, r3, #16
 574:	4670      	mov	r0, lr
 576:	4572      	cmp	r2, lr
 578:	ea6f 0603 	mvn.w	r6, r3
 57c:	bf88      	it	hi
 57e:	4610      	movhi	r0, r2
 580:	eb00 0c06 	add.w	ip, r0, r6
 584:	2001      	movs	r0, #1
 586:	eb00 101c 	add.w	r0, r0, ip, lsr #4
 58a:	f010 0003 	ands.w	r0, r0, #3
 58e:	d00a      	beq.n	5a6 <Reset+0x1a6>
 590:	e891 00f0 	ldmia.w	r1, {r4, r5, r6, r7}
 594:	2801      	cmp	r0, #1
 596:	e883 00f0 	stmia.w	r3, {r4, r5, r6, r7}
 59a:	d130      	bne.n	5fe <Reset+0x1fe>
 59c:	3110      	adds	r1, #16
 59e:	f1bc 0f30 	cmp.w	ip, #48	; 0x30
 5a2:	d204      	bcs.n	5ae <Reset+0x1ae>
 5a4:	e029      	b.n	5fa <Reset+0x1fa>
 5a6:	461a      	mov	r2, r3
 5a8:	f1bc 0f30 	cmp.w	ip, #48	; 0x30
 5ac:	d325      	bcc.n	5fa <Reset+0x1fa>
 5ae:	e891 0089 	ldmia.w	r1, {r0, r3, r7}
 5b2:	e9d1 6503 	ldrd	r6, r5, [r1, #12]
 5b6:	e9d1 4c05 	ldrd	r4, ip, [r1, #20]
 5ba:	f8d1 801c 	ldr.w	r8, [r1, #28]
 5be:	e882 0089 	stmia.w	r2, {r0, r3, r7}
 5c2:	f101 0720 	add.w	r7, r1, #32
 5c6:	e9c2 6503 	strd	r6, r5, [r2, #12]
 5ca:	e9c2 4c05 	strd	r4, ip, [r2, #20]
 5ce:	f8c2 801c 	str.w	r8, [r2, #28]
 5d2:	cf89      	ldmia	r7, {r0, r3, r7}
 5d4:	e9d1 650b 	ldrd	r6, r5, [r1, #44]	; 0x2c
 5d8:	e9d1 4c0d 	ldrd	r4, ip, [r1, #52]	; 0x34
 5dc:	f8d1 803c 	ldr.w	r8, [r1, #60]	; 0x3c
 5e0:	3140      	adds	r1, #64	; 0x40
 5e2:	6210      	str	r0, [r2, #32]
 5e4:	6253      	str	r3, [r2, #36]	; 0x24
 5e6:	6297      	str	r7, [r2, #40]	; 0x28
 5e8:	e9c2 650b 	strd	r6, r5, [r2, #44]	; 0x2c
 5ec:	e9c2 4c0d 	strd	r4, ip, [r2, #52]	; 0x34
 5f0:	f8c2 803c 	str.w	r8, [r2, #60]	; 0x3c
 5f4:	3240      	adds	r2, #64	; 0x40
 5f6:	4572      	cmp	r2, lr
 5f8:	d3d9      	bcc.n	5ae <Reset+0x1ae>
 5fa:	be00      	bkpt	0x0000
 5fc:	e7fe      	b.n	5fc <Reset+0x1fc>
 5fe:	f101 0610 	add.w	r6, r1, #16
 602:	2802      	cmp	r0, #2
 604:	ce74      	ldmia	r6, {r2, r4, r5, r6}
 606:	611a      	str	r2, [r3, #16]
 608:	615c      	str	r4, [r3, #20]
 60a:	619d      	str	r5, [r3, #24]
 60c:	61de      	str	r6, [r3, #28]
 60e:	d10b      	bne.n	628 <Reset+0x228>
 610:	3120      	adds	r1, #32
 612:	f103 0220 	add.w	r2, r3, #32
 616:	f1bc 0f30 	cmp.w	ip, #48	; 0x30
 61a:	d2c8      	bcs.n	5ae <Reset+0x1ae>
 61c:	e7ed      	b.n	5fa <Reset+0x1fa>
 61e:	f104 0108 	add.w	r1, r4, #8
 622:	f100 0308 	add.w	r3, r0, #8
 626:	e79b      	b.n	560 <Reset+0x160>
 628:	f101 0520 	add.w	r5, r1, #32
 62c:	3130      	adds	r1, #48	; 0x30
 62e:	cd35      	ldmia	r5, {r0, r2, r4, r5}
 630:	6218      	str	r0, [r3, #32]
 632:	625a      	str	r2, [r3, #36]	; 0x24
 634:	f103 0230 	add.w	r2, r3, #48	; 0x30
 638:	629c      	str	r4, [r3, #40]	; 0x28
 63a:	62dd      	str	r5, [r3, #44]	; 0x2c
 63c:	f1bc 0f30 	cmp.w	ip, #48	; 0x30
 640:	d2b5      	bcs.n	5ae <Reset+0x1ae>
 642:	e7da      	b.n	5fa <Reset+0x1fa>
```

Possible solutions:

- Recommend using "z". I don't like this solution because optimizations for size penalize inlining and can cause other code to not be properly optimized.

- Implement zero_bss and init_data using assembly making them tight loops like the ones you see in the output of `opt-level = "z"`.

Thoughts @rust-embedded/cortex-m ?

japaric:
before this change a minimal program had 610 bytes of .text
with this change it has 78 bytes of .text

the program was compiled with the default optimizations (opt-level = 3)

closes #110

details in #110

r? @rust-embedded/cortex-m

someone should check that this is not slower and that it actually works


adamgreig:
I'm generally happy with this (and perhaps it will make it easier to add more sections/ramfuncs without too much bloat, too), but does it mean we're now relying on not-stable Rust ABIs?

therealprof:
I'm not happy with this approach. Assembly code and pre-built objects should be a last measure resort. The right thing to do would be to figure out why rust is doing such a non-sense and how to convince it to generate more useful code in the first place.

adamgreig:
> this should be using an extern "C" block.

:+1:

therealprof:
> you mean figure out why LLVM is doing such nonsense. That's not an easy task.

Well, I'd be careful attributing blame here; the problem might still be in rustc.

>  I'd love to use a pure Rust solution but the only solution I can think of is being able to specify optimization levels with item level granularity and that doesn't exist right now.

No, the proper way would be to hint the compiler that the function is called very rarely/exactly once so it optimises it accordingly. In the C/C++ this is extremely common so I'm sure LLVM knows how to handle it.

> I don't agree. If we can fix things today we should fix them today rather than wait for the perfect solution to appear. At the end, users will see the effect of bloat or no bloat and not care much about how it was solved.

This is not a fix, it's a bandaid and exactly covering those two case of initialisation. Same usage patterns inside the application might potentially generate the very same bad code.

> Plus, this approach has the advantage that RAM init will be fast even in unoptimized builds.

For me that is still an assumption.

therealprof:
> I reported this back when LLVM was upgraded (rust-lang/rust#49260). The optimized LLVM IR produced by rustc didn't change; the LLVM upgrade changed the generated machine code.

That still does not mean that rustc produces optimal IR. The parameters (like inlining limits) are still wonky, too.

> How do you hint to the compiler that a function is called very rarely / exactly once in C / C++?

Depends on which compiler and how the code operates, but gcc for instance has `hot` and `cold` attributes (Rustc also seemingly the latter) and some intrinsics like `__builtin_expect` which allow pushing the optimiser in some direction by being less aggressive on code which is known to be less relevant. `__builtin_expect` is mostly used in with macros and e.g. all over the map in the linux kernel to implement `likely`/`unlikely` helping to optimise in a way that a modern CPU can most efficiently chug through.

A compiler a lot of information to make decisions. Small amounts of missing details or incorrect models can easily have an effect like the one you're seeing here. E.g. if the compiler can not make an educated guess about the amount and alignment of data to be copied it might go for a staggered approach with lots of branches to handle the corner cases and then do the bulk of the work with one optimum loop (of course multi-scalar CPUs with deep pipelines and multi level memory/cache architectures have completely different prerequisites than an in-order MCU). Proper annotations can offset that.

However I still think that the reasoning about code generated by a compiler with a default that is equivalent to "make it as fast as possible, no matter the size" is flawed. That's exactly what level 3 means and that is simply IMHO neither appropriate for embedded system nor should we put workaround it's meaning; people get exactly what they ask for.

> Regardless of what you want to call it, this PR reduces the size of all non-Cortex-M0 programs that use cortex-m-rt by ~500 bytes. 

Nicely put. ;)

> Not landing it will negatively impact people's evaluation of Rust for embedded development as this crate is used in the embedded Rust book.

I disagree. Rust is surprisingly resource efficient (except for non-release builds which truly suck, that's something that desperately needs addressing!), ~500 bytes in memory hog mode are not critical, especially when people really should be using size optimisation instead.

> Does this (unoptimized r0::zero_bss) look faster than this (__zero_bss) to you?

Needs a closer look to judge. From my experience I can tell you that the size of a function alone tells precisely nothing about execution speed. Heck, even the project for which you invented the ramfunc veneer interrupt handler stuff I was almost convinced that the version with the twice as long interrupt handler must be slower (it also had a larger amount of branches in it which is often a good indicator of execution speed) but it was almost 60% faster then the shorter version... Sometimes it's also input dependent which version is faster.

I've a suspicion the compiler lost some crucial information like source and/or destination alignment and (non-)overlapping addresses somewhere and thus created code for any possible case.

> We don't have an rfcbot here so I ask you, @therealprof, do you want to register your "this should be fixed in the general case" as a blocking concern for landing this PR?

No, please go ahead an land it if you feel it's important. Let's sure make sure that the real issue (or maybe even two because for me opt level 3 being the default for `release` mode on embedded is somewhat unreasonable) doesn't fall of the radar.

Oh, and one other thing: Usually those `ptr::write` loops are turned into `__aeabi_memcpy` inlines or function calls, how big is `__aeabi_memcpy` on your platform? And is there a way to use it directly from Rust?

korken89:
I think this whole discussion is moot. Currently in Rust there is no way to achieve what we want, hence some method outside of Rust is used.

There is a good reason for RTOSes not using `memset` / `memcpy` in initialization code, they bloat a lot a single use function in the pursuit of speed. Remember that the read / write loop (`.data`) will be 5 clocks where 2 of those clocks are using the memory bus, putting maximum at 2.5x faster by loop unrolling the entire setup. This would however cause unacceptable bloat.

The point here is, this extremely simple way of doing it is close to the maximum we can achieve, and it is the minimal implementation, anyone that is in need of pushing out a faster init only has 3 choices:

1. Cranking up the clock (common and good choice where many times improvement will happen)
2. Loop unrolling to approach the maximum
3. Cranking up the clock and using DMA to maximize utilization of the memory bus (never seen in practice)

With this analysis in mind, the minimal implementation is indeed a sane choice.

therealprof:
> I think this whole discussion is moot. Currently in Rust there is no way to achieve what we want, hence some method outside of Rust is used.

I don't think we have exhausted our options here, yet.

> There is a good reason for RTOSes not using memset / memcpy in initialization code, they bloat a lot a single use function in the pursuit of speed. 

I think you've got that wrong. Most RTOSes I know are jumping into mem* (like Zephyr) or native code similar to what we do in `r0` (like RIOT) and others leave it up the the vendors to supply whatever they want in their BSPs, so in mynewt or mbedOS there's a nightmarish crazy mix of hundreds different both C and ASM custom implementations.

> With this analysis in mind, the minimal implementation is indeed a sane choice.

The current implementation *is* minimal already, it simply is miscompiled into a monster in combination of thumbv7 and default optimisation settings which the assembly is trying to paper over. It does not fix the problem of the compiler generating insane code and by just papering over it (instead of either solving the compilation problem or finding a way to nudge the compiler into not doing that craziness).

korken89:
> I think you've got that wrong. 

Not really, ChibiOS, FreeRTOS would be counter examples. But this are just examples to show for systems which has high usage in user space.

> The current implementation is minimal already

Only from an Rust perspective, not machine language perspective.

> It does not fix the problem of the compiler generating insane code and by just papering over it

It does for our domain's use-case. If you want to fix this in the compiler we need to go through the RFC process to fix a minor pain which can easily be fixed as it is being done here. After all this any one is free to push this issue forward, but for now it needs to be fixed as it gives embedded Rust a view of being "bloat'y".

Plus, we have already showed that there is no real performance to be won in different implementations of this, with the _theoretical_ max of 2.5x improvement, while if you really have issues with startup times the 10x improvement would be the one to chase. The options of clocking up or using DMA is only in the power of the users and should not be done here.

I do not see the issue here, so just to reiterate:

1. We have an implementation which is close to maximum efficiency, there are no 10x improvements to be done
2. It solves the bloat-pain-point
3. It does sadly not use standard Rust, but we do not have a hint for this at this moment, however if ever in the future such a hint is available we can switch to it

therealprof:
@japaric While it helps on thumbv7m-none-eabi, it's a small regression (with default optimisation) on thumbv6m-none-eabi.

 before:

```
# cargo size -- target/thumbv6m-none-eabi/release/examples/main
   text    data     bss     dec     hex filename
     82     192       0     274     112 target/thumbv6m-none-eabi/release/examples/main
```

after PR:
```
# cargo size -- target/thumbv6m-none-eabi/release/examples/main
   text    data     bss     dec     hex filename
     90     192       0     282     11a target/thumbv6m-none-eabi/release/examples/main
```

therealprof:
Just for reference, same application compiled for thumbv7m-none-eabi:

before (opt-level 3):
```
# cargo size -- target/thumbv7m-none-eabi/release/examples/main
   text    data     bss     dec     hex filename
    372    1024       0    1396     574 target/thumbv7m-none-eabi/release/examples/main
```

before (opt-level s):
```
# cargo size -- target/thumbv7m-none-eabi/release/examples/main
   text    data     bss     dec     hex filename
    112    1024       0    1136     470 target/thumbv7m-none-eabi/release/examples/main
```

After PR (opt-level 3 and s):
```
# cargo size -- target/thumbv7m-none-eabi/release/examples/main
   text    data     bss     dec     hex filename
     98    1024       0    1122     462 target/thumbv7m-none-eabi/release/examples/main
```

ithinuel:
I think this is fine to do it in assembler here because we already have HardFault written in assembly.
It's not breaking a "pure rust crate" rule or something.
It's neither introducing risks as it scope is quite small : a private init function.

japaric:
@therealprof

> Usually those ptr::write loops are turned into __aeabi_memcpy inlines or function calls

The way things are implemented at the moment stuff in compiler-builtins like __aeabi_memcpy is *never* inlined into any other code.

The compiler will emit calls to __aeabi_memcpy{,4} for `a.copy_from_slice(b)` depending on alignment if a and b are stack allocated.

Similarly the compiler will emit calls to __aeabi_mem{clr,set}{,4} to initialize large arrays, e.g. `[0u32; 32]`.

> how big is __aeabi_memcpy on your platform?

This symbol is in compiler-builtins. You can objdump the rlib.

``` console
$ arm-none-eabi-objdump -d $(rustc --print sysroot)/lib/rustlib/thumbv6m-none-eabi/lib/libcompiler_builtins-*.rlib
00000000 <__aeabi_memcpy>:
   0:   2a00            cmp     r2, #0
   2:   d005            beq.n   10 <__aeabi_memcpy+0x10>
   4:   780b            ldrb    r3, [r1, #0]
   6:   7003            strb    r3, [r0, #0]
   8:   1c49            adds    r1, r1, #1
   a:   1c40            adds    r0, r0, #1
   c:   1e52            subs    r2, r2, #1
   e:   d1f9            bne.n   4 <__aeabi_memcpy+0x4>
  10:   4770            bx      lr
```

``` console
$ arm-none-eabi-objdump -d $(rustc --print sysroot)/lib/rustlib/thumbv7m-none-eabi/lib/libcompiler_builtins-*.rlib
00000000 <__aeabi_memcpy>:
   0:   b5b0            push    {r4, r5, r7, lr}
   2:   b35a            cbz     r2, 5c <__aeabi_memcpy+0x5c>
   4:   1e53            subs    r3, r2, #1
   6:   f002 0c03       and.w   ip, r2, #3
   a:   2b03            cmp     r3, #3
   c:   d204            bcs.n   18 <__aeabi_memcpy+0x18>
   e:   2200            movs    r2, #0
  10:   f1bc 0f00       cmp.w   ip, #0
  14:   d113            bne.n   3e <__aeabi_memcpy+0x3e>
  16:   e021            b.n     5c <__aeabi_memcpy+0x5c>
  18:   eba2 0e0c       sub.w   lr, r2, ip
  1c:   2200            movs    r2, #0
  1e:   5c8b            ldrb    r3, [r1, r2]
  20:   188c            adds    r4, r1, r2
  22:   5483            strb    r3, [r0, r2]
  24:   1883            adds    r3, r0, r2
  26:   3204            adds    r2, #4
  28:   7865            ldrb    r5, [r4, #1]
  2a:   4596            cmp     lr, r2
  2c:   705d            strb    r5, [r3, #1]
  2e:   78a5            ldrb    r5, [r4, #2]
  30:   709d            strb    r5, [r3, #2]
  32:   78e4            ldrb    r4, [r4, #3]
  34:   70dc            strb    r4, [r3, #3]
  36:   d1f2            bne.n   1e <__aeabi_memcpy+0x1e>
  38:   f1bc 0f00       cmp.w   ip, #0
  3c:   d00e            beq.n   5c <__aeabi_memcpy+0x5c>
  3e:   5c8b            ldrb    r3, [r1, r2]
  40:   f1bc 0f01       cmp.w   ip, #1
  44:   5483            strb    r3, [r0, r2]
  46:   d009            beq.n   5c <__aeabi_memcpy+0x5c>
  48:   1c53            adds    r3, r2, #1
  4a:   f1bc 0f02       cmp.w   ip, #2
  4e:   5ccd            ldrb    r5, [r1, r3]
  50:   54c5            strb    r5, [r0, r3]
  52:   bf08            it      eq
  54:   bdb0            popeq   {r4, r5, r7, pc}
  56:   3202            adds    r2, #2
  58:   5c89            ldrb    r1, [r1, r2]
  5a:   5481            strb    r1, [r0, r2]
  5c:   bdb0            pop     {r4, r5, r7, pc}
```

compiler-builtins and core are shipped as binaries and are always compiled with opt-level = 3.

> And is there a way to use it directly from Rust?

FFI

``` rust
extern "aapcs" {
    fn __aeabi_memcpy(dest: *mut u8, src: *const u8, n: usize);
}

__aeabi_memcpy(..);
```

---

Does loop unrolling ever make sense for Cortex-M devices? I recall that at some
point I compared this unrolled zero_bss vs the tight loop and saw no difference
in execution time with zero Flash latency.

I ask because we could wholesale [disable loop unrolling][no-unroll] for the
Thumb targets. I did a quick test and disabling loop unrolling produces this
with opt-level = 3.

[no-unroll]: https://github.com/rust-lang/rust/blob/0198a1ea45e29af00d92423aa6d2ac876410c3f9/src/librustc_codegen_llvm/back/write.rs#L2218

``` console
$ size -A app
section            size        addr
.text               82        1024

$ arm-none-eabi-objdump -Cd app
00000400 <Reset>:
 400:   f000 f826       bl      450 <DefaultPreInit>
 404:   f240 0000       movw    r0, #0
 408:   f240 0100       movw    r1, #0
 40c:   f2c2 0000       movt    r0, #8192       ; 0x2000
 410:   f2c2 0100       movt    r1, #8192       ; 0x2000
 414:   4281            cmp     r1, r0
 416:   d208            bcs.n   42a <Reset+0x2a>
 418:   f240 0100       movw    r1, #0
 41c:   2200            movs    r2, #0
 41e:   f2c2 0100       movt    r1, #8192       ; 0x2000
 422:   f841 2b04       str.w   r2, [r1], #4
 426:   4281            cmp     r1, r0
 428:   d3fb            bcc.n   422 <Reset+0x22>
 42a:   f240 4064       movw    r0, #1124       ; 0x464
 42e:   f240 0100       movw    r1, #0
 432:   f2c0 0000       movt    r0, #0
 436:   f2c2 0100       movt    r1, #8192       ; 0x2000
 43a:   f850 2b04       ldr.w   r2, [r0], #4
 43e:   f841 2b04       str.w   r2, [r1], #4
 442:   e7fa            b.n     43a <Reset+0x3a>
```

And this is the size with loop unrolling (i.e. unmodified nightly)

``` console
$ size -A app
section            size        addr
.text               546        1024
```

Another data point:

__aeabi_memcpy4 w/o loop unrolling

``` console
00000000 <__aeabi_memcpy4>:
   0:   468c            mov     ip, r1
   2:   4603            mov     r3, r0
   4:   2a04            cmp     r2, #4
   6:   d315            bcc.n   34 <__aeabi_memcpy4+0x34>
   8:   b5b0            push    {r4, r5, r7, lr}
   a:   f1a2 0e04       sub.w   lr, r2, #4
   e:   f02e 0403       bic.w   r4, lr, #3
  12:   1d20            adds    r0, r4, #4
  14:   eb0c 0100       add.w   r1, ip, r0
  18:   4418            add     r0, r3
  1a:   f85c 5b04       ldr.w   r5, [ip], #4
  1e:   3a04            subs    r2, #4
  20:   f843 5b04       str.w   r5, [r3], #4
  24:   2a03            cmp     r2, #3
  26:   d8f8            bhi.n   1a <__aeabi_memcpy4+0x1a>
  28:   ebae 0204       sub.w   r2, lr, r4
  2c:   e8bd 40b0       ldmia.w sp!, {r4, r5, r7, lr}
  30:   f7ff bffe       b.w     0 <__aeabi_memcpy4>
  34:   4661            mov     r1, ip
  36:   4618            mov     r0, r3
  38:   f7ff bffe       b.w     0 <__aeabi_memcpy4>
```

__aeabi_memcpy4 with loop unrolling

``` console
00000000 <__aeabi_memcpy4>:
   0:   2a04            cmp     r2, #4
   2:   bf38            it      cc
   4:   f7ff bffe       bcc.w   0 <__aeabi_memcpy4>
   8:   b5f0            push    {r4, r5, r6, r7, lr}
   a:   1f13            subs    r3, r2, #4
   c:   2401            movs    r4, #1
   e:   f023 0c03       bic.w   ip, r3, #3
  12:   eb04 0493       add.w   r4, r4, r3, lsr #2
  16:   f10c 0e04       add.w   lr, ip, #4
  1a:   f014 0603       ands.w  r6, r4, #3
  1e:   d009            beq.n   34 <__aeabi_memcpy4+0x34>
  20:   460c            mov     r4, r1
  22:   4605            mov     r5, r0
  24:   f854 7b04       ldr.w   r7, [r4], #4
  28:   2e01            cmp     r6, #1
  2a:   f845 7b04       str.w   r7, [r5], #4
  2e:   d104            bne.n   3a <__aeabi_memcpy4+0x3a>
  30:   461a            mov     r2, r3
  32:   e013            b.n     5c <__aeabi_memcpy4+0x5c>
  34:   4605            mov     r5, r0
  36:   460c            mov     r4, r1
  38:   e010            b.n     5c <__aeabi_memcpy4+0x5c>
  3a:   684c            ldr     r4, [r1, #4]
  3c:   2e02            cmp     r6, #2
  3e:   6044            str     r4, [r0, #4]
  40:   d105            bne.n   4e <__aeabi_memcpy4+0x4e>
  42:   3a08            subs    r2, #8
  44:   f101 0408       add.w   r4, r1, #8
  48:   f100 0508       add.w   r5, r0, #8
  4c:   e006            b.n     5c <__aeabi_memcpy4+0x5c>
  4e:   688c            ldr     r4, [r1, #8]
  50:   3a0c            subs    r2, #12
  52:   6084            str     r4, [r0, #8]
  54:   f101 040c       add.w   r4, r1, #12
  58:   f100 050c       add.w   r5, r0, #12
  5c:   4470            add     r0, lr
  5e:   2b0c            cmp     r3, #12
  60:   d30e            bcc.n   80 <__aeabi_memcpy4+0x80>
  62:   6826            ldr     r6, [r4, #0]
  64:   3a10            subs    r2, #16
  66:   602e            str     r6, [r5, #0]
  68:   2a03            cmp     r2, #3
  6a:   6866            ldr     r6, [r4, #4]
  6c:   606e            str     r6, [r5, #4]
  6e:   68a6            ldr     r6, [r4, #8]
  70:   60ae            str     r6, [r5, #8]
  72:   68e6            ldr     r6, [r4, #12]
  74:   f104 0410       add.w   r4, r4, #16
  78:   60ee            str     r6, [r5, #12]
  7a:   f105 0510       add.w   r5, r5, #16
  7e:   d8f0            bhi.n   62 <__aeabi_memcpy4+0x62>
  80:   eba3 020c       sub.w   r2, r3, ip
  84:   4471            add     r1, lr
  86:   e8bd 40f0       ldmia.w sp!, {r4, r5, r6, r7, lr}
  8a:   f7ff bffe       b.w     0 <__aeabi_memcpy4>
```

Would it make sense to disable loop unrolling (for these targets) right now and
in the future ask for a compiler flag to enable loop unrolling?

therealprof:
@japaric 

> Does loop unrolling ever make sense for Cortex-M devices? I recall that at some
point I compared this unrolled zero_bss vs the tight loop and saw no difference
in execution time with zero Flash latency.

There're some very rare circumstances where it might help (Cortex-M3 and higher have speculative branch pre-fetching, Cortex-M7 can have D and I caches) but in the vast majority of cases it's just a horrible waste of flash.

> I ask because we could wholesale disable loop unrolling for the
Thumb targets. I did a quick test and disabling loop unrolling produces this
with opt-level = 3.

That sounds like a better idea to me.

**However** using expensive (in terms of flash size) optimisations is kind of the whole point of opt-level = 3 and by disabling expensive optimisations entirely you're basically degrading opt-level = 3 to opt-level = 2. I still think if somebody absolutely wants to have it, they should be able to get it and instead we should change the default opt-level to a more appropriate value for the vast majority of cases.

Also I wonder what triggers the unrolling in the first place, there must be some incorrect profile setting indicating to the compiler that unrolling this loop is a worthwhile optimisation (but only for thumbv7m and up).

japaric:
@therealprof

> However using expensive (in terms of flash size) optimisations is kind of the
> whole point of opt-level = 3 and by disabling expensive optimisations entirely
> you're basically degrading opt-level = 3 to opt-level = 2.

The aggressive loop unrolling is also enabled in opt-level = 2 AFAICT. The only
differences between 2 and 3 that I can see is that 3 has even higher inlining
threshold than 2 and it enables some extra vectorizer passes.

---

Grepping through clang, rustc and llvm codebases revealed the following:

- In clang loop unrolling can be globally enabled via a codegen flag which is
  [disabled by default][clang]. Clang doesn't seem to enable or disable loop
  unrolling based on the opt level, or any other setting. There are pragmas to
  enable / disable loop unrolling at a fine grained level.

[clang]: https://github.com/llvm-mirror/clang/blob/e978feba76c11c9882c59321a1901445555c1f89/include/clang/Frontend/CodeGenOptions.def#L219

- rustc doesn't seem to either enable or disable loop unrolling in the general
  case. It does disable loop unrolling when opt-level is "s" or "z" -- this
  differs from clang behavior.

- I haven't found definitive proof that LLVM enables loop unrolling by default
  but given that we see loop unrolling in Rust code optimized with opt-level = 2
  that seems to be the case at least for opt-level >= 2.

- However, I did found something interesting in LLVM. For ARM targets LLVM has
  a [loop unrolling preference logic][llvm] that enables more aggressive (?)
  loop unrolling if the target supports the thumb2 instruction set, which is the
  case for the `thumbv7(e)m` targets, but not the `thumbv6m` target. This logic
  appears to have been added 11 months ago. That would explain the differences
  we observe.

[llvm]: https://github.com/llvm-mirror/llvm/blob/master/lib/Target/ARM/ARMTargetTransformInfo.cpp#L580-L582

---

> There're some very rare circumstances where it might help

It seems to me that for these targets the most sensible solution is to disable
loop unrolling by default and have some way to opt into loop unrolling at a fine
grained level (see [#pragma unroll]). A flag to globally enable loop unrolling
would be another option for those that have lots of Flash (and RAM if using
#[ramfunc]) to spare.

[#pragma unroll]: https://clang.llvm.org/docs/AttributeReference.html#pragma-unroll-pragma-nounroll

korken89:
Would disabling loop unrolling remove optimizations such as emitting bit reversal instructions for a loop which does bit reversal manually? I'm no compiler expert.

I would be careful with such a sweeping change unless we have clearly shown that it does what we expect and that this behavior won't change.

japaric:
> Would disabling loop unrolling remove optimizations such as emitting bit reversal instructions for a loop which does bit reversal manually?

Do you have an example where llvm does this optimization? I can check if it still holds with the change. However, such optimizations could break by other reasons like LLVM changing the logic of LLVM passes or just by adding more code around such loop. In general, if you want the compiler to generate an exact instruction you should use the relevant intrinsic. In this case there is a safe (not yet stable) wrapper https://doc.rust-lang.org/std/primitive.u32.html#method.reverse_bits around intrinsics::bitreverse .

korken89:
I don't have a specific example, I was more thinking of loudly from times I have been amazed that loops have been replaced with a few instructions.

If we were to disable loop unrolling, are there any adverse effects predicted or "all is good"?

therealprof:
> However, I did found something interesting in LLVM. For ARM targets LLVM has
a loop unrolling preference logic that enables more aggressive (?)

Actually this previous commit from August 2017 cranked up the unrolling (and also entirely disabled it for Thumb1 for an unknown reason):
https://github.com/llvm-mirror/llvm/commit/f28137119049de0ec44ca53f1e6a62b60b01ce66#diff-062dbf75412335e13b856b4fefdaf033

The rationale provided in the phabricator ticket for doing this in the first place really doesn't make too much sense to me, but as I kind of expected the target here is the view that any branch costs performance (due to additional time required for fetching and executing the branch and to help with branch target forwarding in CM3 and higher: http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.ddi0337e/CACDJFCF.html). So the idea is not actually not to fully unroll the loop but do partial unrolling and/or something we used to call "loop peeling" (not sure that's really used in LLVM lingo). In the end it's all about avoiding flash waitstates caused by additional instruction fetches and executes (the branches and potentially the condition check). However this doesn't take into account that some implementations may not even have waitstates, either because they're running at slow speeds or because they're using quicker flash or because they have something like STs ART implementation which glosses over the WS (and actually might be hindered by excessive unrolling).

> It seems to me that for these targets the most sensible solution is to disable
loop unrolling by default and have some way to opt into loop unrolling at a fine
grained level (see #pragma unroll). A flag to globally enable loop unrolling
would be another option for those that have lots of Flash (and RAM if using
#[ramfunc]) to spare.

I like it.

thejpster:
Forgive my lack of ARM asm knowledge, but as `stmia` does a 32-bit write, does it require the arguments to be 32-bit aligned, as is there anything to ensure the user has configured their linker script that way, or give a nice error (rather than a hard fault) if they haven't?

japaric:
@thejpster we control the linker script in this crate so we [align] the start and end of both .bss and .data to 4 bytes. We also have [assertions] to sanity check that the boundaries are 4-byte aligned. (The previous code also assumed 4-byte aligned; `ptr::{read,write}` assume that the pointer is aligned).

[align]: https://github.com/rust-embedded/cortex-m-rt/blob/581fc9588fe82fc3e5a8856e5c522f5606ae2ea0/link.x.in#L108
[assertions]: https://github.com/rust-embedded/cortex-m-rt/blob/581fc9588fe82fc3e5a8856e5c522f5606ae2ea0/link.x.in#L156-L163

japaric:
Triage: The RAM initialization code can be optimized for size using the `profile-overrides` feature. 

Using:

``` toml
[profile.release.overrides.cortex-m-rt] 
opt-level = "z" 
```

I see a reduction of 488 bytes. I think that feature is more flexible than this PR (though the Cargo feature is unstable atm) so I'm going to close this RFC. If someone feels that we should optimize the RAM initialization for size by default please open a new PR / RFC.

adamgreen:
@japaric  This PR is still referenced in https://docs.rust-embedded.org/book/start/qemu.html#inspecting  Should the documentation be updated to show the `.text` size that will result from an actual build without this PR or should the docs and/or template be updated to include the override noted above?

japaric:
> Should the documentation be updated to show the .text size that will result from an actual build without this PR

Yes

> should the docs and/or template be updated to include the override noted above?

The template should compile out of the box on stable. If it includes the override it should be commented out.

I would propose that:

- we include the original disassembly (no profile override) but with the unrolled loop replaced by `(..)` to keep the disassembly short
- include a comment saying that loop unrolling is enabled on opt-level >=2 and that produces a large binary that has 2x speed up over the unrolled loop.
- mention that the loop unrolling can be disabled without changing the opt-level using a profile override

Rua:
I came here noticing that the message is still present in the book. I guess nothing has been done since 2018?

PR #114, opened 2018-09-09
v0.6.3 and v0.1.1

japaric:
fixes the `rand` issues


therealprof:
Works. Just published new fresh version of the microbit crate with all the latest bells and whistles (and hardware random number support).

Issue #109, opened 2018-09-07
How to define interrupt handlers?


therealprof:
Now that we have the fantastic new `#[entry]` and `#[exception]` macros, how do we deal with interrupt handlers?

japaric:
Ideas:

1. When the "device" feature is not enabled and cortex-m-rt registers 32 (v6-M) or 240 (v7-M) interrupts in the vector table we could have an `#[interrupt]` attribute that uses *position* rather than name:

``` rust
// strawman syntax
#[interrupt(0)] // first interrupt handler
fn foo() {}

#[interrupt(1)] // second interrupt handler
fn bar() {}

#[interrupt(240)] // this is an error
fn baz() {}
```

2. When the "device" feature is enabled and the device crate provides N named interrupt we change the syntax of the `#[interrupt]` attribute to:

``` rust
#[interrupt]
fn USART1() {}
```

To check that USART1 is a valid interrupt name we could (ab)use the fact that attributes are *not* in the type namespace to have e.g. `use stm32f103xx::interrupt` import *both* the attribute and an enumeration of valid interrupt names. Then the expansion of `#[interrupt]` would use the imported interrupt to check if it's valid or not. With code:

``` rust
// crate: stm32f103xx

extern crate cortex_m_rt;

pub use cortex_m_rt::_interrupt as interrupt; // this is the attribute

// this is the enumeration of valid interrupt names
enum interrupt { USART1, USART2, .. }
```

Then this

``` rust
// user code

use stm32f103xx::interrupt; // imports both the attribute and the enumeration

#[interrupt]
fn USART1() {
    // code goes here
}
```

would expand into

``` rust
// expanded user code

use stm32f103xx::interrupt;

#[export_name = "USART1"]
pub extern "C" fn some_random_hash() {
    interrupt::USART1;  // checks that USART1 exists in the enumeration

    // code goes here
}
```

The downside of this hack is that this wouldn't work:

``` rust
#[stm32f103xx::interrupt]
fn USART1() {
    // code goes here
}
```

because the interrupt enum has not been imported. Though I don't think it's idiomatic to use attributes without importing them first.

I haven't actually tested that this last part about the check works but I think it should work.

Also, we don't necessarily have to do (1) *and* (2). We could do only (2).

therealprof:
[...]

PR #122, opened 2018-09-17
[RFC] `#[interrupt]`

japaric:
this PR adds an `interrupt` attribute that's similar to the existing `exception`
except that it's used to override device-specific interrupt handlers.

The other big difference is that `cortex_m_rt::interrupt` can't be directly
used; it must first be imported from a device crate, which re-exports
`cortex_m_rt::interrupt`. This is required to check at compile time that the
interrupt is valid for the target device.

Safe `static mut` variables can be used with `#[interrupt]` handlers. The syntax
is the same as `#[exception]`s.

More background information can be found in #109.

The companion svd2rust PR is rust-embedded/svd2rust#235

r? @rust-embedded/cortex-m

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ cortex-m-rt-macros = { path = "macros", version = "0.1.1" }
 cortex-m = "0.5.4"
 panic-abort = "0.3.0"
 panic-semihosting = "0.4.0"
+panic-halt = "0.2.0"
 
 [dev-dependencies.rand]
 default-features = false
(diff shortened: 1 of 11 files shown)


japaric:
> Why did you change all examples to panic_halt?

Oops. I meant to sent that as a separate PR. See #124. I'll remove that commit from this PR.

adamgreig:
I've pushed a merge commit to resolve the merge conflicts. Could someone have a quick check and approve, then we'll merge this?

adamgreig:
bors retry

adamgreig:
You can just put "static mut bla: u32 = 0" in the top of your interrupt function, and the macro transforms it to an `&mut` which you can safely use in the rest of the function.

adamgreig:
It works the same as the `#[exception]` macro: https://docs.rs/cortex-m-rt-macros/0.1.3/cortex_m_rt_macros/attr.exception.html#examples

birkenfeld:
Ah, nice! That's quite nifty :)

PR #124, opened 2018-09-17
use panic-halt instead of panic-{abort,semihosting}

japaric:
the former requires a feature gate; the later pulls in the cortex-m crate

Eventually we want to test this crate on beta. To do that our dependencies must
not contain feature gates. That's not the case of `panic-abort`, which depends
on the `core_intrinsics` feature.

This PR fixes that. I could have used `panic-semihosting` everywhere but decided
for `panic-halt` which has zero dependencies as I figured that might slightly
decrease CI times.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,8 +16,7 @@ cortex-m-rt-macros = { path = "macros", version = "0.1.1" }
 
 [dev-dependencies]
 cortex-m = "0.5.4"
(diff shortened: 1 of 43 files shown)


adamgreig:
I think `examples/unsafety.rs` still uses `panic_semihosting`?

japaric:
@adamgreig updated

adamgreig:
bors retry

Issue #125, opened 2018-09-18
Regression on 0.6 on rust messages with attribute macro


TeXitoi:
Hi!

When using the new attribute macro (`#[entry] fn main() -> ! { loop {} }`, I have a big regression when compiling and using a debugger: everything (error, warning and debugger) point to the attribute line, making debugging tricky.

A workaround is to directly call another function, but that's quite dirty:
```rust
#[entry]
fn entry() -> ! {
    main()
}
fn main() -> ! {
    loop {}
}
```

adamgreig:
This is caused by [#90 (comment)](https://github.com/rust-embedded/cortex-m-rt/pull/90#issuecomment-414086570), in other words a problem in proc-macro2. I think the hope is proc-macro2 will land the required feature soon.

TeXitoi:
See https://github.com/alexcrichton/proc-macro2/issues/104 and https://github.com/rust-lang/rust/issues/43081 for reference.

TeXitoi:
Maybe updating to syn 0.15 will fix this problem: https://github.com/dtolnay/syn/releases/tag/0.15.0

japaric:
@TeXitoi do you have steps to reproduce the problem you are seeing? That way we can evaluate if moving to syn 0.15.0 helps or not.

in particular, I'm seeing that debugging works as intended on 1.30-beta. This is using the dev profile.

![20180919-213505](https://user-images.githubusercontent.com/5018213/45776761-3064fe80-bc54-11e8-869b-0971d1844152.png)

What version of GDB are you using?

Compile errors on statements also have the correct span.

``` rust
// ..

#[entry]
fn main() -> ! {
    let x: () = 42;

    loop {}
}
```

``` console
$ cargo build
error[E0308]: mismatched types
  --> src/main.rs:14:17
   |
14 |     let x: () = 42;
   |                 ^^ expected (), found integral variable
   |
   = note: expected type `()`
              found type `{integer}`
```

Where I see incorrect spans is when e.g. the `#[entry]` function doesn't type check.

``` rust
// ..

#[entry]
fn main() -> ! {}
```

```
$ cargo build
error[E0308]: mismatched types
 --> src/main.rs:9:1
  |
9 | #[entry]
  | ^^^^^^^^ expected !, found ()
  |
  = note: expected type `!`
             found type `()`
```

---

> This is caused by #90 (comment)

Enabling the nightly feature of proc-macro2 should be documented in the crate level documentation.

[...]

PR #127, opened 2018-09-19
bump the syn dependency

japaric:
and switch to the recommended way to parse tokens: `parse_macro_input!`.

This improves (?) error messages when the user applies one of our attributes to
an item that's not a function.

Consider

``` rust
 #[entry]
static MAIN: () = ();

```

The error message changed from:

```
error: custom attribute panicked
  --> src/main.rs:10:1
   |
10 | #[entry]
   | ^^^^^^^^
   |
   = help: message: `#[entry]` must be applied to a function: ParseError(Some("failed to parse fn item: failed to parse"))
```

to:

```
error: expected `fn`
  --> src/main.rs:11:1
   |
11 | static MAIN: () = ();
   | ^^^^^^

```

---

Before landing this I'd like to hear more details about #125 to see if this
helps

### DIFF
--- a/macros/Cargo.toml
+++ b/macros/Cargo.toml
@@ -13,12 +13,12 @@ version = "0.1.1"
 proc-macro = true
 
 [dependencies]
-quote = "0.6.6"
-proc-macro2 = "0.4.15"
+quote = "0.6.8"
+proc-macro2 = "0.4.19"
 
 [dependencies.syn]
 features = ["extra-traits", "full"]
-version = "0.14.8"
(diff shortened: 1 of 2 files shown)


japaric:
Applying `needs-decision` label just to prevent this from being merged until we have more details on issue #125 .

japaric:
Looks like we won't be able to get more information from issue #125 so we can go ahead and land this.

thejpster:
I can't see anything wrong, but I've never used `syn` and I don't know what I'm looking at.

PR #134, opened 2018-10-23
device example only with device cfg

therealprof:


therealprof:
Argh... Let's try this again...

PR #135, opened 2018-10-23
Only compile device example if "device" feature is activated

therealprof:


adamgreig:
> in v0.6.4 and before.

I think v0.6.3 was the last one that could build all examples.

therealprof:
I bisected it (between v0.6.4 and master) and it came up with this:
```
# git bisect good
24cb97188d7f93f12e6171b545369a7f8739e4d9 is the first bad commit
```

PR #140, opened 2018-10-26
entry/exception/interrupt: improvements to the `static mut` transformation

japaric:
see individual commits for details


therealprof:
Interesting, I was wondering why there was a reason to do that indirectly but forgot to ask. Seems there's not. 👍

PR #143, opened 2018-10-26
`b UserHardFault`

japaric:
this commit replaces the `bl UserHardFault` instruction in HardFault with `b
UserHardFault`. This lets us drop the prologue (`push {r0,lr}`) while preserving
the ability to unwind the stack using GDB.

To prevent linker errors about relocations when UserHardFault and HardFault end
up far away from each other and the target is ARMv6-M (where the `b` instruction
only supports offsets of +/- 2KB) we use two *input* sections: .HardFault and
.UserHardFault. HardFault is placed in the .HardFault section and UserHardFault
is placed in the .UserHardFault section. The .HardFault input section is placed
in the output .text section after all the input .text sections, and the
.UserHardFault input section is placed after the .HardFault section. This
causes the two symbols to always appear next to each other in the output binary,
furthermore UserHardFault *always* appears after HardFault so the branch offset
of `b UserHardFault` is always a few bytes.

It should be noted that neither .HardFault or .UserHardFault will appear in the
output of the `size -A` command as they are input sections that get merged into
the output .text section. IOW, the sizes of HardFault and UserHardFault will
continue to be reported under the .text section.

cc @adamgreen


adamgreen:
@japaric Looks good to me.

PR #144, opened 2018-10-27
[RFC] rename UserHardFault to HardFault

japaric:
so it matches the exception name (`#[exception] fn HardFault(..`)

Right now the symbol name of all exception handlers match the name of the
function used with the `#[exception]` attribute *except* for `HardFault`, whose
symbol name actually is `UserHardFault`.

This PR corrects that inconsistency by renaming the `UserHardFault` symbol to
`HardFault`, and the `HardFault` symbol to `HardFaultTrampoline`.

This change doesn't break compilation or changes functionality but it does soft
break GDB scripts that include the command `break UserHardFault` (e.g. the GDB
script in cortex-m-quickstart) in the sense that the breakpoint will no longer
work. However the rest of the GDB script will continue to work.

RFC questions: (a) do we want to do this rename? (b) if the answer is yes, do we
want to include this rename in the v0.5.x release, or should we consider it a
breaking change and postpone it until v0.6.0?


thejpster:
I'm happy for this to go out as a v0.5.x.

adamgreig:
Likewise happy for it to be in 0.5.x.

adamgreig:
@rust-embedded/cortex-m I've fixed the failing build; please re-review.

japaric:
triage: this has 4 approvals (majority). Changing labels to decision-accepted and S-waiting-on-review.

adamgreig:
bors r=korken89

japaric:
rebased / force pushed to hopefully unstuck bors

japaric:
latest commit should fix the [build error](https://travis-ci.org/rust-embedded/cortex-m-rt/jobs/454769783)

adamgreig:
Looks like your rebase got rid of _both_ of my commits, including the one updating the expected error message in the compile-fail test?

RandomInsano:
Hey guys, this seems to have broken the book because of how [Semver versioning works](https://semver.org/):

https://github.com/rust-embedded/book/issues/113

So, the question is should the version of cortext-m-rt have been bumped to 1.0.0 for this as it broke the book? Also, should I re-open that issue?

PR #147, opened 2018-11-03
add thumbv8m baseline

evq:


therealprof:
@evq I don't know what the status of thumbv8m as a target in Rust is but we once that is rolling we need to add it to CI.

evq:
I have a PR open to add the baseline target - which is blocked on another PR for a cargo change:
https://github.com/rust-lang/rust/pull/55041

therealprof:
@evq Maybe open tickets so we don't forget?

therealprof:
@evq Hm, you used a different toolchain to build this it seems.

therealprof:
```
+diff -u bin/thumbv8m.base-none-eabi.after bin/thumbv8m.base-none-eabi.before
--- bin/thumbv8m.base-none-eabi.after	2018-11-03 23:01:03.168136948 +0000
+++ bin/thumbv8m.base-none-eabi.before	2018-11-03 23:01:03.120136241 +0000
@@ -3,8 +3,8 @@
 cortex-m-rt.o:     file format elf32-littlearm
 
 
-Disassembly of section .HardFault:
+Disassembly of section .text.HardFault:
 
 00000000 <HardFault>:
    0:	f3ef 8008 	mrs	r0, MSP
-   4:	e7fe      	b.n	0 <UserHardFault>
+   4:	f7ff fffe 	bl	0 <UserHardFault>
```

evq:
Ah, I'm using `arm-none-eabi-binutils-2.31.1-1` on Arch

evq:
Looks like the version installed in `ci/install.sh` is slightly older?
```
GNU assembler (GNU Tools for Arm Embedded Processors 7-2018-q2-update) 2.30.0.20180329
```

therealprof:
I guess more important is the question: Which one is the better version.

evq:
```
+   4:	f7ff fffe 	bl	0 <UserHardFault>
```

I think this is correct - right?

Since `asm.s` uses the `bl` instruction:
```
  .section .text.HardFault
  .global HardFault
  .thumb_func
HardFault:
  mrs r0, MSP
  bl UserHardFault
```

The ARMv8-M architecture reference manual seems to indicate that the `bl` instruction has two 16-bit components - which seems to match

adamgreig:
You might not have an up to date master - asm.s recently (a week ago)
changed to use b instead of bl.

evq:
aha, thanks @adamgreig, that'll do it! :)

PR #155, opened 2018-12-06
ci: test on stable

japaric:
now that 1.31 is out


japaric:
an error message changed in a nightly upgrade. The latest commit fixes the compile-fail test.

re-r? @therealprof

PR #167, opened 2019-01-16
Add thumbv8m main.

nezza:
rustup supports the `thumbv8m.main-none-eabi` target, this adds support for `thumbv8m.main-none-eabi` to cortex-m-rt.

Related PRs:

https://github.com/rust-embedded/cortex-m-semihosting/pull/28
https://github.com/rust-embedded/cortex-m/pull/130


nezza:
The Travis build is failing because it relies on cortex-m supporting thumbv8m.main already - after the PR in cortex-m is integrated it should work.

therealprof:
bors try

nezza:
@adamgreig mentioned we have to wait for a new release, cause the cortex-m module is version fixed

thejpster:
Can I do anything to push this along? I have an M33 and I want to try it out!

therealprof:
bors try

mathk:
Ping from triage: @nezza will resolving iusee https://github.com/rust-embedded/cortex-m/issues/131 make bors happy ?

nezza:
My understanding is that a cortex-m (minor) release would make bors happy (someone in IRC said so at least :) )

nezza:
bors try

adamgreig:
bors try

thejpster:
thumv8m support requires cortex-m 0.6+ I believe. Also it seems the assembler files don't match.

nezza:
Yep, I'll push an increase to 0.6.0 in a bit

thejpster:
You may also need a newer GCC. I'm having trouble with the arm-none-eabi-gcc included with Ubuntu 18.04. The latest pre-built GCC from ARM (gcc-arm-none-eabi-8-2018-q4-major) works fine.

nezza:
Thanks for the hint, will take a look!

adamgreig:
The build is already using cortex-m 0.6 (since I updated Cargo.toml to use >0.5.7 <0.7), but it's _also_ using cortex-m 0.5.7 for something else I suppose. Anyway the failure in this case seems to be the assembler files.

You can see it's using cortex-m 0.6 here: https://travis-ci.org/rust-embedded/cortex-m-rt/jobs/506709530
and you can see which version of the GCC toolchain it's using here: https://github.com/rust-embedded/cortex-m-rt/blob/master/ci/install.sh#L10

thejpster:
https://crates.io/crates/cortex-m-semihosting hasn't been updated, so that's pulling in cortex-m 0.5.x. Not sure why this crate depends on -semihosting.

mathk:
@rust-embedded/cortex-m seems that we have 2 PR for the same feature ? #182

mathk:
@thejpster looks like it is in [example/qemu.rs](https://github.com/rust-embedded/cortex-m-rt/blob/master/examples/qemu.rs)

thejpster:
> @rust-embedded/cortex-m seems that we have 2 PR for the same feature ? #182

Mine also adds the missing cfg flags. You can do this one then mine, or just mine.

korken89:
Merged #182 and I think this one should not be needed, if I missed something reopen this. 
Thanks for all the work!

PR #168, opened 2019-01-20
On HardFault entry use MSP/PSP depending on stack mode

bytex64:
Noticed this when making my own hardfault handler - none of the registers were correct. The `&ExceptionFrame` argument was wrong because I was using the PSP in Thread mode, but the handler only loads MSP. This PR inspects the EXC_RETURN value in LR and loads either MSP or PSP depending on the stack used (see section B1.5.8 in the ARMv7-M architecture manual for details on the format of EXC_RETURN).

I haven't tested this on armv6m or armv8m, but I believe it works the same across the board. Also didn't update the libraries in `bin/` since I wasn't sure if that was expected or not.


adamgreig:
You need to run `assemble.sh` yourself and commit the resulting objects. The CI checks the committed objects match what it builds itself and fails the build if they don't.

bytex64:
Roger that. I've compiled and added the libs. I'm using the Gnu ARM Embedded Toolchain 8-2018-q4-major if that makes any difference.

thejpster:
(Checked against https://www.pjrc.com/teensy/beta/DDI0403D_arm_architecture_v7m_reference_manual.pdf and yup, bit 4 indicates main/process stack).

adamgreen:
👍  This matches how I have done similar MSP/PSP checking in debug monitor code.

korken89:
bors r=thejpster

PR #173, opened 2019-01-25
Fix linker asserts for vector table not located at address 0.

hugwijst:
Force `__reset_vector` and `__eexceptions` symbols to be absolute in linker script assert.

At least `arm-none-eabi-ld` will interpret the symbol location as relative to the section they were defined in. This caused the asserts to fail when `.vector_table` is not placed at memory address 0.

### DIFF
--- a/link.x.in
+++ b/link.x.in
@@ -177,10 +177,10 @@ BUG(cortex-m-rt): start of .heap is not 4-byte aligned");
 /* # Position checks */
 
 /* ## .vector_table */
-ASSERT(__reset_vector == ADDR(.vector_table) + 0x8, "
+ASSERT(ABSOLUTE(__reset_vector) == ADDR(.vector_table) + 0x8, "
 BUG(cortex-m-rt): the reset vector is missing");
 
-ASSERT(__eexceptions == ADDR(.vector_table) + 0x40, "
+ASSERT(ABSOLUTE(__eexceptions) == ADDR(.vector_table) + 0x40, "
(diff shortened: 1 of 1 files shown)


korken89:
Sorry for the late reply here, do you have a project which shows this fix in effect?

hugwijst:
Not publicly :)

You can reproduce this by changing the start location of `FLASH` in `memory.x` to a non-0 value (eg. 0x1000) and building any example using the GNU linker (eg. `cargo rustc --target thumbv7em-none-eabihf --example main -- -C linker=arm-none-eabi-ld`).

I'm running into this problem because I want to integrate a project with a bootloader, which requires the vector table to be at a non-standard location.

adamgreig:
I think there must be something else going on here, since many projects use `FLASH` starting at non-zero (e.g. any STM32 has flash origin at 0x0800_0000) and they build fine using GNU linker. Specifically your example also succeeds without error for me. What version of arm-none-eabi-ld are you using?

I've also built projects which use a bootloader and so the application has the flash origin offset by a sector or two without issue, which also seem to build fine with GNU gcc or ld.

hugwijst:
Closing PR as this seems to be a linker bug.

I ran into this with the `arm-none-eabi-ld` distributed with Ubuntu 18.04:
```GNU ld (2.27-9ubuntu1+9) 2.27```

I don't see this issue with either 2.27.90 (6-2017-q1-update) and 2.29.51 (7-2017-q4-major) from https://developer.arm.com/open-source/gnu-toolchain/gnu-rm/downloads.

PR #179, opened 2019-03-02
Fix incorrect __EXCEPTIONS in lib.rs comment

raz-a:
Update lib.rs comment describing __INTERRUPTS variable to say __INTERRUPTS instead of __EXCEPTIONS


adamgreen:
Nice catch!

PR #182, opened 2019-03-12
Add thumbv8m.main support.

thejpster:
* Add thumbv8m.main support.
* Also adds feature flags into build.rs so SecureFault gets included.


thejpster:
Ok, so there are two issues:

1) There's some overlap with #167. That's fine, I can re-base.
2) I have no idea how many Interrupt Vectors to set as the maximum when building for a non-Cortex-M target (which we do why?)

therealprof:
@thejpster previously we just used the same number as for non-armv6m. But the explicit `panic!` also doesn't help a bit... ;)

adamgreig:
Hopefully this (or #167, however that works out) can be merged once #183 is done.

adamgreig:
bors try

adamgreig:
@thejpster looks like you have some more-useful errors to investigate now ;)

thejpster:
Builds are now green.

PR #185, opened 2019-03-18
Remove redundant and incorrect target FPU check

jonas-schievink:
Most target triples with FPU end in `gnueabihf` or `musleabihf`, so this isn't catching them. Since the `has_fpu` function does, this shouldn't change behavior.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -73,10 +73,6 @@ handlers.");
         max_int_handlers
     ).unwrap();
 
-    if target.ends_with("-eabihf") {
-        println!("cargo:rustc-cfg=has_fpu");
-    }
-
     println!("cargo:rustc-link-search={}", out.display());
 
     println!("cargo:rerun-if-changed=build.rs");


thejpster:
I'm a little confused by this. I thought cortex-m-rt only supported bare-metal no_std thumbv[678]e?m-none-eabi(hf)? So it doesn't matter if it can't parse gnueabi and musl targets?

jonas-schievink:
Yes, that's true, sorry for the confusion. There might be more targets added in the future which use a different ABI, but for now this just removes redundant code.

thejpster:
How else would the 'has_fpu' feature get set, which is used in src/lib.rs to gate whether to include the FPU-enable code?

jonas-schievink:
There's a `has_fpu` function futher down in the build script. It's called at the start of the build script.

thejpster:
Oh yes. I was confused by the two identical pieces of code, and how "View File" in Github opens the file in your branch (i.e. post-changes).

PR #118, opened 2018-09-17
[RFC] Keep `.stack_sizes`

japaric:
PR [rust-lang/rust#51946] will add a `-Z emit-stack-sizes` to `rustc` that can
be used to (make LLVM) emit stack usage metadata. Most linkers will discard this
metadata by default, however.

[rust-lang/rust#51946]: https://github.com/rust-lang/rust/pull/51946

This PR / RFC proposes that we modify our linker script to keep this metadata
and place in an output `.stack_sizes` section (the canonical section name that
LLVM uses for this metadata).

This `.stack_sizes` section will be marked as `(INFO)` meaning that it won't be
loaded into the Flash or RAM of the target device.

The advantage of doing this is that tools that parse this information will work
out of the box. See examples at the bottom.

Not doing this means that users will have to mess with linker scripts and tweak
the linking process to be able to access the stack usage metadata. See [this
example] to get an idea of the required steps to keep the metadata information.

[this example]: https://github.com/japaric/stack-sizes#example-usage

# Examples

Using the [`cargo stack-sizes`] subcommand to build a project and print its
stack usage information. Assumes that this PR has been merged.

[`cargo stack-sizes`]: https://github.com/japaric/stack-sizes

``` console
$ cargo stack-sizes --bin app -v
RUSTC=stack-sizes-rustc "cargo" "rustc" "--bin" "app" "--"
    Finished dev [unoptimized + debuginfo] target(s) in 0.01s
"stack-sizes" "target/thumbv7m-none-eabi/debug/app"
address         stack   name
0x00000416      112     <cortex_m_rt::ExceptionFrame as core::fmt::Debug>::fmt::h4362577979112554
0x0000059e      96      <<cortex_m_rt::ExceptionFrame as core::fmt::Debug>::fmt::Hex as core::fmt::Debug>::fmt::hd24e21694d63c2ec
0x0000069c      72      core::fmt::Arguments::new_v1_formatted::hc22bfa5fcec35f0e
0x00000758      48      r0::init_data::hdc91f9605f364aeb
0x00000710      40      r0::zero_bss::h24ed73bae17eec88
0x000007d8      24      core::ptr::<impl *mut T>::offset::h940561014e707589
0x000007fc      24      core::ptr::<impl *const T>::offset::h6847f0bc0fe1f0b5
0x00000820      24      core::ptr::read::ha83613479c1b8688
0x00000858      24      core::sync::atomic::compiler_fence::h89dd3725007a5019
[...]

PR #186, opened 2019-03-24
do not KEEP the .stack_sizes section

japaric:
this undoes PR #118

I found a problem with keeping this section. Turns out that the input
`.stack_sizes` sections contain references to all functions compiled with `-Z
emit-stack-sizes` (the section contains the addresses of all those functions
after all) so keeping those section prevents the linker from removing *any* of
those functions. This is not a problem today because `-Z emit-stack-sizes` is
*opt-in* and only used to analyze a program.

However, I am proposing a rust-lang/rust PR to build the `compiler-builtins`
crate with `-Z emit-stack-sizes`. That change will cause *all* the functions in
that crate to be kept in binaries that link to `cortex-m-rt`, regardless of
whether the crate author uses `-Z emit-stack-sizes` or not, leading a increase
in binary size of about 14 KB (`.text` section).

To prevent issues with that rust-lang/rust PR I propose we undo PR #118. On the
bright side, the tools that were depending on this (`cargo-stack-sizes` and
`cargo-call-stack`) no longer do so in their latest releases so nothing is lost
on the tooling front with this change.

r? @rust-embedded/cortex-m

### DIFF
--- a/link.x.in
+++ b/link.x.in
@@ -131,12 +131,6 @@ SECTIONS
   /* Place the heap right after `.bss` */
   __sheap = ADDR(.bss) + SIZEOF(.bss);
 
-  /* Stack usage metadata emitted by LLVM */
-  .stack_sizes (INFO) :
-  {
-    KEEP(*(.stack_sizes));
-  }
-
   /* ## .got */
   /* Dynamic relocations are unsupported. This section is only used to detect relocatable code in
      the input files and raise an error if relocatable code is found */


japaric:
> However, I am proposing a rust-lang/rust PR

PR in question: rust-lang/rust#59401

japaric:
compiletest v0.3.19 doesn't compile with the latest nightly. We can either wait or mark the nightly build as `allow_fail: true`

therealprof:
@japaric I'm feeling that is the only right thing to do. Using CI on nightly is a total gamble and waste of time.

korken89:
@japaric Rebase with master to get CI to pass

japaric:
@korken89 there are no merge conflicts; bors should work

bors r=therealprof

PR #190, opened 2019-05-31
Align sections

### LINKED ISSUE #189: .bss is not 4-byte aligned
In [this sx1276 project](https://github.com/lthiery/sx1276-rs), I'm trying to link to C code compiled using the 'cc' crate (by the way, thanks for the comment about the `fPIC` flag). 

I was able to exercise some of the functions, but as I continued to build it out, I ran into the following error:
```
          BUG(cortex-m-rt): .bss is not 4-byte aligned
          
          rust-lld: error: 
          BUG(cortex-m-rt): start of .heap is not 4-byte aligned
```

I assume my static C-library has brought something in that misaligns things, but I'm confused how the following doesn't more or less guarantee alignment:

```
  /* ### .bss */
  .bss : ALIGN(4)
  {
    *(.bss .bss.*);

    . = ALIGN(4); /* 4-byte align the end (VMA) of this section */
  } > RAM
```
I've tried adding `__attribute__ ((aligned (8)))` to no avail. 

Wonder if anyone has ideas? Hack fixes welcome as much as nice fixes :)

NickeZ:
Hey, I had the same issue as in #189. Don't ask me why this works I looked at other linker scripts.

Fixes #189

### DIFF
--- a/link.x.in
+++ b/link.x.in
@@ -105,14 +105,13 @@ SECTIONS
   /* ### .data */
   .data : ALIGN(4)
   {
+    . = ALIGN(4);
+    __sdata = .;
     *(.data .data.*);
-
     . = ALIGN(4); /* 4-byte align the end (VMA) of this section */
+    __edata = .;
   } > RAM AT > FLASH
 
-  /* VMA of .data */
-  __sdata = ADDR(.data);
-  __edata = ADDR(.data) + SIZEOF(.data);
 
   /* LMA of .data */
   __sidata = LOADADDR(.data);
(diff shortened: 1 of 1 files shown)


thejpster:
I'll see if I can create a reproducer for #189 and test this out later today.

therealprof:
The change looks good to me but I'm really curious why this is necessary. Maybe some crappy/picky linkers in play?

NickeZ:
I saw this technique in the template linker scripts you get from microchip (atmel). You can see an example here: https://github.com/digitalbitbox/mcu/blob/master/bootloader.ld

I'm not using any strange linker, it is GCC 8.

https://developer.arm.com/-/media/Files/downloads/gnu-rm/8-2018q4/gcc-arm-none-eabi-8-2018-q4-major-linux.tar.bz2?revision=d830f9dd-cd4f-406d-8672-cca9210dd220?product=GNU%20Arm%20Embedded%20Toolchain,64-bit,,Linux,8-2018-q4-major

lthiery:
This fixed my issue. I don't understand it either, but it works :+1:

korken89:
I had a look at the ChibiOS linker implementation, which also does this trick. It seems as the first align should be enough, but multiple sources double do it. I was also not able to reproduce though.
Feels like some linker trickery, but adding this should not alter the original intention so I will accept this.

PR #193, opened 2019-06-24
Feature request: allow custom memory sections in the linker script

WRansohoff:
It looks like the Rust Cortex-M linker script currently expects a `memory.x` file to define the chip's memory regions and a `device.x` file to define weak aliases for each interrupt handler, but I don't see a way to define custom memory sections.

From what I can tell, the `SECTIONS { ... }` part of the linker script assumes that there will only be two main memory regions called `RAM` and `FLASH`, and it places common memory sections like `text`, `bss`, `data`, etc. into those sections.

But some chips have multiple non-contiguous memory banks, and it is not uncommon for applications to split up contiguous memory regions to ensure that important blocks such as caches, heaps, large constant arrays, etc. are placed correctly. For example, you might assign a filesystem cache to a fast block of core-coupled RAM, or you might assign all of your application's constant audio/video data to a known area of Flash memory which the application can configure as read-only.

So, it would be nice if the linker script accepted an optional `sections.x` file where you can define mappings for extra memory sections. I think that this pull request should work, but my 'best effort' testing was not comprehensive and feedback is very welcome.

In order to make the `sections.x` file optional, I added a few lines to the build script which create an empty file with the same name. Suggestions for a better way to do this are welcome, but you can override the empty `sections.x` file using the same sort of build logic that the `cortex_m_quickstart` project uses to populate the `memory.x` file:

    let out_path = PathBuf::from( env::var( "OUT_DIR" ).unwrap() );
    macro_rules! ld_sec { () => { "path/to/your_sections.x" }; };
    File::create( out_path.join( "sections.x" ) )
        .expect( "Failed to open memory sections file" )
        .write_all( include_bytes!( ld_sec!() ) )
        .expect( "Failed to write memory sections file" );
    println!( "cargo:rustc-link-search={}", out_path.display() );

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -42,6 +42,10 @@ INCLUDE device.x"#
         f
     };
 
+    // Write an empty `sections.x` file to be overwritten by the application if desired.
+    let mut f_sections = File::create(out.join("sections.x")).unwrap();
+    f_sections.write(b"\n").unwrap();
+
     let max_int_handlers = if target.starts_with("thumbv6m-") {
         println!("cargo:rustc-cfg=cortex_m");
         println!("cargo:rustc-cfg=armv6m");
--- a/link.x.in
+++ b/link.x.in
@@ -145,6 +145,9 @@ SECTIONS
     *(.ARM.exidx.*);
     *(.ARM.extab.*);
   }
+
+  /* User-defined memory sections */
+  INCLUDE sections.x
 }
 
 /* Do not exceed this mark in the error messages below                                    | */
--- a/sections.x
+++ b/sections.x
@@ -1,0 +1,8 @@
+/* Sample sections.x file */
+/* Ensure at least 1KB of RAM is left for heap space. */
+.heap_guard :
+{
+  . = ALIGN(4);
+  . = . + 1K;
+  . = ALIGN(4);
+} > RAM


build.rs
```diff
@@ -42,6 +42,10 @@ INCLUDE device.x"#
         f
     };
 
+    // Write an empty `sections.x` file to be overwritten by the application if desired.
+    let mut f_sections = File::create(out.join("sections.x")).unwrap();
+    f_sections.write(b"\n").unwrap();
+
```
Disasm: Are you sure it works? I investigated this a couple of months ago and found that there is no guaranteed order for search directories. With custom `sections.x` you can get either it or this empty `sections.x` included.

adamgreig:
I'm away from my computer right now but until I can write a full response,
note that this is already supported as you can have multiple `SECTIONS`
blocks so just appending them to your `memory.x` file works and has been
used before. You can even specify how to concatenate them. For example from
my FFP project's `memory.x` ([here](
https://github.com/adamgreig/ffp/blob/master/firmware/memory.x)):

```
MEMORY
{
    FLASH  (rx) : ORIGIN = 0x08000000, LENGTH = 32K
    RAM    (rw) : ORIGIN = 0x20000000, LENGTH = 6K
    USBRAM (rw) : ORIGIN = 0x40006000, LENGTH = 1K
}

SECTIONS
{
    .usbram (NOLOAD) : ALIGN(4)
    {
        *(.usbram .usbram.*);
        . = ALIGN(4);
    } > USBRAM
} INSERT AFTER .bss;
```

You can then [use them](
https://github.com/adamgreig/ffp/blob/master/firmware/src/hal/usb/buffers.rs#L26)
with `#[link_section]`, but beware they won't get automatically loaded by
r0 (unlike bss and statics). That's also true of the extra sections you
define here, though. There's an open RFC for more comprehensive support
including r0 initialisation [#164](
https://github.com/rust-embedded/cortex-m-rt/issues/164), but there's a lot
more work to get there and it's not yet clear exactly what it would look
like or how it would work.

I haven't had time to go through this PR in full yet but I _think_ being
able to define extra sections in the existing memory.x provides the same
features?

WRansohoff:
Oh, thank you - I wasn't aware that the linker would combine multiple `SECTIONS` blocks. I think that does obviate the need for this, and it's probably a better solution given what @Disasm said about the search paths not having a guaranteed order.

I'll just close this, then. Thanks!

PR #195, opened 2019-07-20
Update rand to 0.7.0

rubdos:
Since a bit, `rand` has `small_rng` as a separate feature. I thus had to add it as required feature.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,7 +37,8 @@ panic-halt = "0.2.0"
 
 [dev-dependencies.rand]
 default-features = false
-version = "0.5.5"
+version = "0.7.0"
+features = ["small_rng"]
 
 [target.'cfg(not(target_os = "none"))'.dev-dependencies]
 compiletest_rs = "0.3.14"
--- a/macros/Cargo.toml
+++ b/macros/Cargo.toml
@@ -21,8 +21,8 @@ features = ["extra-traits", "full"]
 version = "0.15.13"
 
 [dependencies.rand]
-version = "0.5.5"
-default-features = false
+version = "0.7.0"
+features = ["small_rng"]
 
 [dev-dependencies]
 cortex-m-rt = { path = "..", version = "0.6.0" }


rubdos:
There's still [an issue](https://github.com/rust-random/rand/pull/844) with no_std in `rand` itself.

therealprof:
@rubdos Indeed. I also noticed that while trying to upgrade some other crates of mine a while ago. Just curious: What's the benefit of this version bump?

rubdos:
> @rubdos Indeed. I also noticed that while trying to upgrade some other crates of mine a while ago. Just curious: What's the benefit of this version bump?

Fair point... I was bumping it because I was trying to get futures + async_await to compile, and I noticed that `futures-preview` contained 0.7.0. Without the version bump, I'd have multiple versions of the same crate...

It's nothing urgent though, I don't even have my test boards yet :D

therealprof:
@rubdos The use of rand in this crate is somewhat special and also a bit unfortunate, bumping it to the latest version sounds good to me in general, but it would be even better if we could completely eliminate it. 😅

rubdos:
Ah, tbqf, I didn't look at why it needed `rand` at all :sweat: 

It's not lost energy though, since that issue I found in `rand` itself applies to any `no_std` environment, and `futures-preview` uses it too if you enable `async-await`.

I've had a look at the use, and it's just to generate some unique identifiers with a bunch of macros? I don't get why it'd need to be `no_std` compatible then, if it's presumably only used on compile time, and not even included in the output?

therealprof:
@rubdos You can only have one set of dependencies with one set of feature-flags (feature flags are always additive) in your application. So it's normally only used at build time, but in case you want to use it in an application you're stuck with the same one...

rubdos:
> @rubdos You can only have one set of dependencies with one set of feature-flags (feature flags are always additive) in your application. So it's normally only used at build time, but in case you want to use it in an application you're stuck with the same one...

I'd agree, but in this case, there's one `rand` in the macro's library, and another `rand` in the root crate. Both were at different versions (and probably features too?) before 7e31e50.

If I remove `rand` from the main `Cargo.toml`, everything except the example still compiles, afaict.

therealprof:
> If I remove rand from the main Cargo.toml, everything except the example still compiles, afaict.

Sure.

PR #196, opened 2019-07-23
Update rand to 0.7.0

timothyhollabaugh:
This updates the `rand` crate to the latest version 0.7.0.

I've been working on https://github.com/timothyhollabaugh/micromouse, which includes a cargo workspace with both a no_std bin crate for an stm32f4 (`firmware`) and a std bin crate (`simulation`) for running on a normal computer. One of the dependencies in `simulation` depends on `rand` 0.6.5, and  `firmware` depends on `cortex-m-rt-macros` which depends on `rand` 0.5.5. However, due to [issue 645 in the `rand` repo](https://github.com/rust-random/rand/issues/645), having both these versions of `rand` with one being no_std and one std in the dependency tree causes compile errors. Hence the upgrade to 0.7.0.

Everything builds, and I have tested it with the uart isr in the above repo.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,7 +37,7 @@ panic-halt = "0.2.0"
 
 [dev-dependencies.rand]
 default-features = false
-version = "0.5.5"
+version = "0.7.0"
 
 [target.'cfg(not(target_os = "none"))'.dev-dependencies]
 compiletest_rs = "0.3.14"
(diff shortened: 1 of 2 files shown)


therealprof:
Hi @timothyhollabaugh, there's already #195 which does the same thing. But there's a problem in `rand´...

timothyhollabaugh:
I probably should've looked at the PRs before diving in...

I can wait until #195 gets resolved. In the meantime, I suppose you can close this if you want.

PR #202, opened 2019-08-19
Enable building semihosting for Armv8m

aurabindo:
Semihosting protocol has not changed for Arvmv8m and hence it
need not be disabled for this architecture.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,21 +19,10 @@ autoexamples = true
 r0 = "0.2.2"
 cortex-m-rt-macros = { path = "macros", version = "0.1.5" }
 
-[target.thumbv7em-none-eabihf.dev-dependencies]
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -19,21 +19,10 @@ autoexamples = true
...
-[target.thumbv7m-none-eabi.dev-dependencies]
-cortex-m-semihosting = "0.3.1"
-
-[target.thumbv6m-none-eabi.dev-dependencies]
-cortex-m-semihosting = "0.3.1"
-
 [dev-dependencies]
 cortex-m = "0.6"
 panic-halt = "0.2.0"
+cortex-m-semihosting = "0.3"
```
aurabindo: @korken89 Is this change fine?

PR #205, opened 2019-09-27
Stop using randomized symbol names

### LINKED ISSUE #212: `static mut` transform allows aliasing `&mut` references
PoC:
```rust
#![feature(asm)]
#![no_std]
#![no_main]

extern crate cortex_m_rt as rt;
extern crate panic_halt;
extern crate cortex_m;
extern crate cortex_m_semihosting as semi;

extern crate stm32l0xx_hal;

use rt::{entry, exception};
use cortex_m::interrupt::{self, Mutex};
use core::cell::RefCell;
use semi::hprintln;

static STORAGE: Mutex<RefCell<Option<&'static mut u8>>> = Mutex::new(RefCell::new(None));

#[exception]
fn SVCall() {
    static mut STORED: bool = false;
    static mut STAT: u8 = 0;

    let stat: &'static mut u8 = STAT;

    if *STORED {
        let alias = interrupt::free(move |cs| {
            STORAGE.borrow(cs).borrow_mut().take().unwrap()
        });

        hprintln!("stat ={:p}", stat).unwrap();
        hprintln!("alias={:p}", alias).unwrap();
        loop {
            unsafe { asm!(""); }
        }
    } else {
        interrupt::free(move |cs| {
            *STORAGE.borrow(cs).borrow_mut() = Some(stat);
        });
        *STORED = true;
    }
}

#[entry]
fn you_died_of_dis_entry() -> ! {
    hprintln!("hello").unwrap();
    loop {
        unsafe { asm!("svc 0"); }
    }
}
```

Output:

```
hello
stat =0x20000009
alias=0x20000009
```

jonas-schievink:
It isn't possible to do this by incrementing a global counter, since the expansion order of macros isn't guaranteed and might change between compiler invocations.

Fixes #212
Closes https://github.com/rust-embedded/cortex-m-rt/pull/196
Closes https://github.com/rust-embedded/cortex-m-rt/pull/195


therealprof:
The randomisation was added on purpose so we should discuss if we prefer to not have it. Maybe @japaric would like to weigh in on this?

korken89:
The reason for the hash was to make it ~impossible to guess the name of the internal functions, however I'd argue that better reproducibility in builds is much more important.

japaric:
This change (making the name predictable) makes the crate unsound. Example:

``` rust
#[entry]
fn main() -> ! {
    loop {
        __cortex_m_rt_internal_SysTick(); // predictable and SAFE to call
        // BUT this call may be preempted by `SysTick` which can lead to
        // two &mut- references to `X` to exist in the stack frame === data race
    }
}

#[exception]
fn SysTick() {
    static mut X: u64 = 0;
    // ..
}
```

If you want to drop the name randomization then one option is a trampoline:

``` rust
#[exception]
fn SysTick() {
    static mut X: u64 = 0;
    // ..
}
```

Expansion:

``` rust
// UNSAFE -- because it creates a reference to a `static mut` from thin air
#[no_mangle]
unsafe extern "C" SysTick() {
    __cortex_m_rt_internal_SysTick({
        static mut X: u64 = 0;
        &mut X
    })
}

// SAFE -- note that the argument has _non_ 'static lifetime
fn __cortex_m_rt_internal_SysTick(X: &mut u64) {
    // ..
}
```

There are other options; just make sure that the function that creates a reference to `static mut` variables is unsafe to call.

jonas-schievink:
> This change (making the name predictable) makes the crate unsound.

Sure, if you consider the generated code part of the cortex-m-rt's public API. I don't think it should be though (hence the `__cortex_m_rt_internal_` prefix).

I will explore the trampoline option. If it's possible to make it completely safe without compromising determinism then that's even better, of course (provided that the implementation doesn't become massively complicated, but it shouldn't).

korken89:
Has there been any progress here? I am a bit interested in how this pans out as we saw the issue first hand with not having reproducible builds in our firmware.

jonas-schievink:
Okay, this should now be ready for review.

jonas-schievink:
Agreed on both counts.

therealprof:
@japaric Looks good to me but it would be great if you could give it another set of eyes since you're much more familiar with the implementation and suggested the change.

jonas-schievink:
compiletest-rs is extremely fragile when there's existing build artifacts, so I've pushed a commit that disables the Travis cache. Hopefully that fixes it.

PR #211, opened 2019-11-01
Bump rand dependency to 0.7

therealprof:


therealprof:
Both #195 and #196 are missing different tiny details, so third time is a charm. ;)

Issue #138, opened 2018-10-24
Links to macros broken (on docs.rs at least)


birkenfeld:
The three links to the macros in the cortex-m-rt-macros crate [here](https://docs.rs/cortex-m-rt/0.6.5/cortex_m_rt/) are broken: they lead to e.g. https://docs.rs/cortex-m-rt/0.6.5/cortex_m_rt_macros/fn.entry.html

adamgreig:
Good spot. I'm not sure if there's a good way to cause the correct links to be generated, given they'll have to contain the version number of the cortex-m-rt-macros crate (i.e., https://docs.rs/cortex-m-rt-macros/0.1.3/cortex_m_rt_macros/attr.exception.html)? The public re-exports at the bottom of the page does have correct links for the macros.

jonas-schievink:
Hmm, I think we could put `#[doc(inline)]` on the reexport and treat `cortex-m-rt-macros` as an unstable implementation detail that users shouldn't use directly.

jonas-schievink:
Huh, looks like the current version of the macros crate doesn't even work on docs.rs: https://docs.rs/cortex-m-rt-macros/

PR #219, opened 2019-11-26
Inline attr. macro docs and fix links

jonas-schievink:
This makes `cortex-m-rt-macros` purely an implementation detail, so I've put a warning in place to reflect that.

This should close https://github.com/rust-embedded/cortex-m-rt/issues/138


korken89:
bors r=therealprof

PR #221, opened 2019-12-04
Version 0.6.11

jonas-schievink:


therealprof:
We should update the CHANGELOG.md before we publish. Care to create another PR?

PR #222, opened 2019-12-04
Bump the changelog for 0.6.11

jonas-schievink:
Unfortunately I already published it. Should we publish 0.6.12 just for this fix?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -468,7 +468,7 @@ section                size        addr
 
 Initial release
 
-[Unreleased]: https://github.com/rust-embedded/cortex-m-rt/compare/v0.6.10...HEAD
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -467,6 +469,7 @@ section                size        addr
 Initial release
 
 [Unreleased]: https://github.com/rust-embedded/cortex-m-rt/compare/v0.6.10...HEAD
```
adamgreig: This line also needs updating to 0.6.11

therealprof:
I guess we can live with the slight mismatch of CHANGELOG.md and the release.

adamgreig:
> Should we publish 0.6.12 just for this fix?

I don't think we need to, the changelog is presumably very rarely viewed from the released crate.

therealprof:
Would be great to have the tag on the right version though...

PR #227, opened 2019-12-28
Hide trampolines from docs

### LINKED ISSUE #226: #[interrupt] shows trampoline in docs
While looking at #225 I just noticed that the generated `__cortex_m_rt_TIMER0_trampoline` function shows up in the docs. Given that this isn't really a public function it seems like it should have `#[doc(hidden)]` applied to it.

korken89:
Closes #226


korken89:
bors r=jonas-schievink

Issue #213, opened 2019-11-21
Multiple attributes are allowed on a single function


jonas-schievink:
...leading to incorrect error messages:

```rust
#[exception]
#[entry]
fn SVCall() -> ! {
    loop {}
}
```
```
error: `#[entry]` function must have signature `[unsafe] fn() -> !`
  --> examples/bug.rs:17:1
   |
17 | #[exception]
   | ^^^^^^^^^^^^
```

The function already has the signature that the error message demands, and the span points at the *other* attribute instead of the function.

PR #228, opened 2019-12-28
Added so attributes and cfgs are applied on the trampolines as well

### LINKED ISSUE #223: `#[link_section]` (and maybe more) should be applied to the trampoline
In some not-yet-open-sourced code, we have used `#[link_section]` to place 2 functions, one of which is an `#[exception]` handler, in the same section so they're closer together, in order to link correctly (one of them calls the other using inline assembly with limited jump range).

With https://github.com/rust-embedded/cortex-m-rt/pull/205, this code broke because `#[link_section]` does not get applied to the trampoline, and the compiler decides to inline the function *with* the `#[link_section]` attribute into the trampoline. Whether this inlining decision makes sense is another question, but this still broke code that one should reasonably expect to work.

There's a few ways to fix this or work around the issue:

* Put `#[inline(never)]` on the user code (this is what I've done for now, and it fixes linkage, but wastes code space because this really *should* be inlined)
* Declare the inlining as a rustc bug, report it, and forget about it.
* Have the `#[exception]` macro (and the others) duplicate some attributes and apply them to both function in the output.

The last solution would be ideal, since it fixes the user-visible regression in behavior without preventing the inlining, and makes attributes on `#[exception]` functions behave more similarly to how they behave on regular, single functions. I don't know which *other* linkage-related attributes need to be duplicated though.

korken89:
I made a first go at #223, but maybe this implementation takes things we do not want to propagate.
If you come up with something that should not propagate, please comment here.

Closes  #223
Closes https://github.com/rust-embedded/cortex-m-rt/issues/213

### DIFF
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -839,7 +839,7 @@ fn check_attr_whitelist(attrs: &[Attribute]) -> Result<(), TokenStream> {
                 return Err(
                     parse::Error::new(
                         attr.span(),
-                        "cortex-m-rt does not support multiple attributes on the same function",
+                        "multiple cortex-m-rt attributes are not supported on the same function",
                     )
                     .to_compile_error()
                     .into(),
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -150,7 +150,15 @@ pub fn entry(args: TokenStream, input: TokenStream) -> TokenStream {
         })
         .collect::<Vec<_>>();
 
+    if let Some(error) = check_for_blacklisted_attrs(&f.attrs) {
+        return error;
+    }
+
+    let (ref cfgs, ref attrs) = extract_cfgs(f.attrs.clone());
+
     quote!(
+        #(#cfgs)*
+        #(#attrs)*
         #[doc(hidden)]
         #[export_name = "main"]
         pub unsafe extern "C" fn #tramp_ident() {
@@ -343,7 +351,15 @@ pub fn exception(args: TokenStream, input: TokenStream) -> TokenStream {
             let tramp_ident = Ident::new(&format!("{}_trampoline", f.sig.ident), Span::call_site());
             let ident = &f.sig.ident;
 
+            if let Some(error) = check_for_blacklisted_attrs(&f.attrs) {
+                return error;
+            }
+
+            let (ref cfgs, ref attrs) = extract_cfgs(f.attrs.clone());
+
             quote!(
+                #(#cfgs)*
+                #(#attrs)*
                 #[doc(hidden)]
                 #[export_name = #ident_s]
                 pub unsafe extern "C" fn #tramp_ident() {
@@ -396,7 +412,15 @@ pub fn exception(args: TokenStream, input: TokenStream) -> TokenStream {
             let tramp_ident = Ident::new(&format!("{}_trampoline", f.sig.ident), Span::call_site());
             let ident = &f.sig.ident;
 
+            if let Some(error) = check_for_blacklisted_attrs(&f.attrs) {
+                return error;
+            }
+
+            let (ref cfgs, ref attrs) = extract_cfgs(f.attrs.clone());
+
             quote!(
+                #(#cfgs)*
+                #(#attrs)*
                 #[doc(hidden)]
                 #[export_name = "HardFault"]
                 #[link_section = ".HardFault.user"]
@@ -481,7 +505,15 @@ pub fn exception(args: TokenStream, input: TokenStream) -> TokenStream {
                 })
                 .collect::<Vec<_>>();
 
+            if let Some(error) = check_for_blacklisted_attrs(&f.attrs) {
+                return error;
+            }
+
+            let (ref cfgs, ref attrs) = extract_cfgs(f.attrs.clone());
+
             quote!(
+                #(#cfgs)*
+                #(#attrs)*
                 #[doc(hidden)]
                 #[export_name = #ident_s]
                 pub unsafe extern "C" fn #tramp_ident() {
@@ -650,7 +682,15 @@ pub fn interrupt(args: TokenStream, input: TokenStream) -> TokenStream {
         })
         .collect::<Vec<_>>();
 
+    if let Some(error) = check_for_blacklisted_attrs(&f.attrs) {
+        return error;
+    }
+
+    let (ref cfgs, ref attrs) = extract_cfgs(f.attrs.clone());
+
     quote!(
+        #(#cfgs)*
+        #(#attrs)*
         #[doc(hidden)]
         #[export_name = #ident_s]
         pub unsafe extern "C" fn #tramp_ident() {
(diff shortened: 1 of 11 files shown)


korken89:
bors try

jonas-schievink:
This will also apply attributes like `#[export_name]` and `#[inline]`, which doesn't make too much sense to me. I think the most robust approach would be to use a whitelist of attributes that are propagated to the trampoline. I'll start working on one.

korken89:
Sure, should be no problem to add.
But not a blacklist? Seems like `doc`, `inline`, and `export_name` are the direct ones we do not want.

korken89:
@jonas-schievink I added a blacklist, works as expected. Are there more we would like to blacklist?
Or do you have a specific whitelist I should change to?

macros/src/lib.rs
```diff
@@ -790,6 +840,20 @@ fn extract_cfgs(attrs: Vec<Attribute>) -> (Vec<Attribute>, Vec<Attribute>) {
     (cfgs, not_cfgs)
 }
 
+fn containts_blacklist_attrs(attrs: &Vec<Attribute>) -> bool {
```
jonas-schievink:
```suggestion
fn contains_blacklist_attrs(attrs: &[Attribute]) -> bool {
```

macros/src/lib.rs
```diff
@@ -396,7 +416,17 @@ pub fn exception(args: TokenStream, input: TokenStream) -> TokenStream {
             let tramp_ident = Ident::new(&format!("{}_trampoline", f.sig.ident), Span::call_site());
             let ident = &f.sig.ident;
 
+            let (ref cfgs, ref attrs) = extract_cfgs(f.attrs.clone());
+
+            if containts_blacklist_attrs(cfgs) || containts_blacklist_attrs(attrs) {
+                return parse::Error::new(Span::call_site(), "This attribute is not allowed")
```
jonas-schievink:
```suggestion
                return parse::Error::new(Span::call_site(), "this attribute is not allowed")
```

This should either point at the attribute in question, or mention the attribute in the error message. Right now it will point at the cortex-m-rt macro.

korken89: Fixed this

korken89:
Fixed your comments.

> The blacklist approach should also handle `#[cfg_attr(some_cfg, blacklisted_attr)]`.

This is already handled and tested for :)

korken89:
Also, I will rebase and squash this before we merge this

korken89:
Tests added to CI as well

korken89:
There, I think this should be ready now

thejpster:
Handing to @jonas-schievink as this is a bit out of my comfort zone.

macros/src/lib.rs
```diff
@@ -790,6 +830,30 @@ fn extract_cfgs(attrs: Vec<Attribute>) -> (Vec<Attribute>, Vec<Attribute>) {
     (cfgs, not_cfgs)
 }
 
+fn check_for_blacklisted_attrs(attrs: &[Attribute]) -> Option<TokenStream> {
+    if let Some(val) = containts_blacklist_attrs(attrs) {
+        return Some(parse::Error::new(val.span(), "This attribute is not allowed [blacklisted]")
```
jonas-schievink:
```suggestion
        return Some(parse::Error::new(val.span(), "this attribute is not allowed on a function controlled by cortex-m-rt")
```

For extra points, mention the exact type of cortex-m-rt attribute here (exception handler, interrupt handler, entry point).

korken89:
Sure, we can switch to a whitelist as well.
What more than `doc` and `link_section` should be allowed?

korken89:
Updated per your comments, ready for another review

macros/src/lib.rs
```diff
@@ -351,7 +351,7 @@ pub fn exception(args: TokenStream, input: TokenStream) -> TokenStream {
             let tramp_ident = Ident::new(&format!("{}_trampoline", f.sig.ident), Span::call_site());
             let ident = &f.sig.ident;
 
-            if let Some(error) = check_for_blacklisted_attrs(&f.attrs) {
+            if let Err(error) = check_attr_whitelist(&f.attrs) {
```
jonas-schievink: Does this need to be duplicated into every match arm or can be move before the match?

korken89: Not sure I follow what you are asking here. Could you elaborate?

jonas-schievink: `exception` has 3 cases (`DefaultHandler`, `HardFault`, and "everything else"), and right now they all call `check_attr_whitelist`. I think it would be less duplication if there was just one call to it at the beginning of the `exception` function.

korken89: Ah right, fixed it

jonas-schievink:
Oh one more thing, does this happen to also fix https://github.com/rust-embedded/cortex-m-rt/issues/213 by any chance? I'm not sure if proc. macro attributes can see other proc. macro attributes, but if so this should fail the whitelist check (and if not that issue can be closed as wontfix).

korken89:
Sure! I added a test for that and it does catch this issue as well.

korken89:
Hmm, I should probably look into a better error messages for these

korken89:
I will have a look at the message, should go fast

korken89:
@jonas-schievink I added better diagnostics, please have a look again and then we can merge

macros/src/lib.rs
```diff
@@ -790,6 +823,44 @@ fn extract_cfgs(attrs: Vec<Attribute>) -> (Vec<Attribute>, Vec<Attribute>) {
...
+                continue 'o;
+            }
+        }
+
+        for val in cortex_m_rt_blacklist {
+            if eq(&attr, &val) {
+                return Err(
+                    parse::Error::new(
+                        attr.span(),
+                        "cortex-m-rt does not support multiple attributes on the same function",
```
jonas-schievink: Hmm, this wording suggests that *no* other attributes are ever allowed, but the whitelisted ones are. What about this wording instead:

```suggestion
                        "multiple cortex-m-rt attributes are not supported on the same function",
```

korken89:
Sounds good, I committed it

macros/src/lib.rs
```diff
@@ -790,6 +823,44 @@ fn extract_cfgs(attrs: Vec<Attribute>) -> (Vec<Attribute>, Vec<Attribute>) {
     (cfgs, not_cfgs)
 }
 
+fn check_attr_whitelist(attrs: &[Attribute]) -> Result<(), TokenStream> {
+    let whitelist = &["doc", "link_section", "cfg", "allow", "warn", "deny", "forbid", "cold"];
+    let cortex_m_rt_blacklist = &["entry", "exception", "interrupt", "pre_init"];
```
jonas-schievink: One issue with this is that the attributes can also be used fully-qualified (`#[cortex_m_rt::exception]`) and `#[interrupt]` is reexported by PACs, so using `#[stm32l0x2::interrupt]` also works and isn't detected here.

This isn't a terribly big issue since they still fail the whitelist check, but it's something that can potentially be improved. One way to do that would be to pass the type of macro we're currently expanding to the `check_attr_whitelist` function (entry, exception, interrupt, or pre_init), and incorporate that into the error message. It could then look like:

> this attribute is not allowed on an exception handler
> this attribute is not allowed on an interrupt handler
> this attribute is not allowed on a cortex-m-rt entry point
> this attribute is not allowed on a pre-init function

These would work regardless of whether they point to a non-whitelist attribute or a cortex-m-rt one.

We can also just do this later of course. Thoughts?

korken89: I added so error messages are better now

macros/src/lib.rs
```diff
@@ -836,27 +855,34 @@ fn check_attr_whitelist(attrs: &[Attribute]) -> Result<(), TokenStream> {
...
+                    WhiteListCaller::Exception => {
+                        &"this attribute is not allowed on an exception handler"
+                    }
+                    WhiteListCaller::Interrupt => {
+                        &"this attribute is not allowed on an interrupt handler"
+                    }
+                    WhiteListCaller::PreInit => {
+                        &"this attribute is not allowed on an interrupt handler"
+                    }
+                };
```
jonas-schievink: These messages can be used below as well. In fact, the blacklist can be removed now (or the issue with fully-qualified attribute paths would remain).

macros/src/lib.rs
```diff
@@ -836,27 +855,34 @@ fn check_attr_whitelist(attrs: &[Attribute]) -> Result<(), TokenStream> {
...
+
+                return Err(parse::Error::new(attr.span(), err_str)
```
jonas-schievink: Can do `&err_str` here and remove the `&`s in front of the string literals above

korken89:
Very true, updated and removed blacklist

PR #229, opened 2019-12-28
Fixed changed compile-fail messages

korken89:
Fixed the errors now present in `compile-fail` tests.


korken89:
This will cause CI to pass on #228 and #227

jonas-schievink:
bors +r

PR #239, opened 2020-01-11
Make `ExceptionFrame`s fields private

### LINKED ISSUE #234: Make fields of `ExceptionFrame` private
We might want `ExceptionFrame` to support accessing FPU registers in the future. Adding these would be a breaking change, since all of its fields are public.

When writing an RTOS, it may be useful to be able to modify the stacked registers, which is generally wildly unsafe, but could be still done correctly. To do this, a `&mut ExceptionFrame` needs to be passed to the handler. With the current API this is immediately unsound, since the code in the handler can set them without requiring `unsafe`.

Making the fields private and having safe getters and unsafe setters would allow this sort of API to be provided.

jonas-schievink:
Closes #234

~~I can also add the `unsafe` setters, but they don't have any use right now.~~ (added them in order to not regress available operations on `ExceptionFrame`)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -416,33 +416,60 @@ pub use macros::{entry, exception, pre_init};
 #[doc(hidden)]
 pub static __ONCE__: () = ();
 
-/// Registers stacked (pushed into the stack) during an exception
+/// Registers stacked (pushed into the stack) during an exception.
 #[derive(Clone, Copy)]
 #[repr(C)]
 pub struct ExceptionFrame {
-    /// (General purpose) Register 0
-    pub r0: u32,
+    r0: u32,
+    r1: u32,
+    r2: u32,
+    r3: u32,
+    r12: u32,
+    lr: u32,
+    pc: u32,
+    xpsr: u32,
+}
 
-    /// (General purpose) Register 1
-    pub r1: u32,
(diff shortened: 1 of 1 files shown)


Disasm:
Is there any sense in modifying only some of the full set of registers? If you implement an RTOS, you may want to save/restore all of the registers.

jonas-schievink:
> Is there any sense in modifying only some of the full set of registers? If you implement an RTOS, you may want to save/restore all of the registers.

Yes, I've had to set only `pc` in RTOS code, to jump into userspace once the interrupt returns. Also, for `SVCall`, you generally want to read the registers that pass arguments to the syscall, and then write back the result (which can end up in the same set of registers, or only `r0`). In my case these registers were `r0-3`.

Saving / Restoring all registers isn't a bad point though, since that's needed for context switching. Maybe we can provide wrappers around pointer casts / transmute / `ptr::{read, write}` that can be used to deal with entire `ExceptionFrame`s at a time? I'm open to suggestions here.

(this reminds me that none of these functions are `#[inline]`, I should definitely fix that before this is merged)

jonas-schievink:
bors r-

inline attributes are still missing

jonas-schievink:
bors r=korken89

PR #243, opened 2020-01-26
Run CI on all branches

jonas-schievink:
This is useful for testing branches of old versions that are still maintained (eg. `v0.6.x` here).

Maintainers are encouraged to open PRs from their forks instead of branches in this repo to prevent CI from running twice. I think switching to GitHub Actions would also prevent this if I'm not mistaken (Travis is a bit weird here, there's no point in running the same pipeline twice).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,12 +63,6 @@ script:
 
 after_script: set +e
 
-branches:
-  only:
-    - master
-    - staging
-    - trying
-
 notifications:
   email:
     on_success: never


adamgreig:
Does `bors try` not work for such testing?

jonas-schievink:
Yeah, looks like it does work. It's still be nicer to have CI run immediately, but I'm also fine with not merging this.

adamgreig:
The current setup comes from the main WG [ops notes](https://github.com/rust-embedded/wg/blob/master/ops/post-transfer.md) and I think is the bors recommendation for Travis, so I'd rather keep it as-is if we can, or work out changing what we do wg-wide, rather than special casing cortex-m-rt. I don't see any strong reason to not run CI on all branches in particular, though.

jonas-schievink:
Okay, that makes sense. Closing this for now. I don't expect too many backports anyways.

PR #246, opened 2020-02-05
Fix doctests and run them in CI

jonas-schievink:
This moves them from the macros crate to the base crate, avoiding a "weird cyclic dev-dependency". It also makes all of them buildable, and does that in CI.


therealprof:
Very nice!

Issue #253, opened 2020-03-19
Better defaults for DefaultHander and Hardfault


korken89:
@jamesmunns and I had a discussion on Matrix about the current defaults that leave `HardFault_` and `DefaultHander_` implemented as infinite loops.
This should most likely be replaced with `panic` in the default case to help the implementors detect issues and if nothing is done the panic crate takes care of the default course of action.
This might increase code size a bit, but for those that do not want this can override it and provide their own implementation.

From an API standpoint it is not a breaking change, however the documentation states that we use infinite loops as the default, so it is a breaking change.

jonas-schievink:
An alternative would be to panic when `#[cfg(debug_assertions)]` is on, and use infinite loops otherwise. That way users won't have to go through the hassle of overriding handlers to get small executables.

jamesmunns:
Could do, but honestly I'd prefer exceptions to reset if I've selected `panic-reset` or `panic-persist` as my panic handler, even in release mode.

If you have a panic handler that disregards the panicinfo, it should still be largely optimized out (branch to panic handler/`SCB::sys_reset()`, instead of infinite branch loop)

jonas-schievink:
This should probably have been closed by #257, but that has since been reverted, with the decision to make overriding non-maskable interrupt handlers `unsafe` (https://github.com/rust-embedded/cortex-m-rt/pull/289), so closing as wontfix instead.

PR #257, opened 2020-03-30
Changed Hardfault's and DefaultHander's default implementations to panic

korken89:
From the discussions in #253 I started a PR with the proposed change.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -207,7 +207,7 @@
 //! trampoline is required to set up the pointer to the stacked exception frame.
 //!
 //! - `HardFault`. This is the user defined hard fault handler. If not overridden using
-//! `#[exception] fn HardFault(..` it will default to a panic with message "Hardfault".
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -945,21 +945,17 @@ pub unsafe extern "C" fn Reset() -> ! {
 #[link_section = ".HardFault.default"]
 #[no_mangle]
 pub unsafe extern "C" fn HardFault_(ef: &ExceptionFrame) -> ! {
-    loop {
-        // add some side effect to prevent this from turning into a UDF instruction
-        // see rust-lang/rust#28728 for details
-        atomic::compiler_fence(Ordering::SeqCst);
-    }
+    panic!("Hardfault");
```
adamgreig: Nit-picking here but could this be `"HardFault"`? Documentation above would have to change too.

korken89: Good catch, changed to this

PR #207, opened 2019-10-22
Rejig link.x to include more lables to help the linker lay out the ob…

therealprof:
…jects

Tested with 1.34.0 and 1.38.0 and careful inspection of the linker map
generated on the previously failing
https://github.com/rust-lang/rust/files/3722440/minimal-rust-lld-issue.zip

Closes #188 (I believe)
Closes https://github.com/rust-lang/rust/issues/65391

Signed-off-by: Daniel Egger <daniel@eggers-club.de>

PR #260, opened 2020-04-03
Update linker script to fix arm-none-eabi-gcc linking. Closes #259

### LINKED ISSUE #259: arm-none-eabi-gcc broken with current link.x.in
In #207 we updated the linker script to fix #188 but unfortunately this seems to have broken arm-none-eabi-gcc, which can no longer link:

```
section .init LMA [0000000008001614,000000000800161f] overlaps section .rodata LMA [0000000008001614,00000000080017cf]
```
The exact problem is adding `__extext` here:
https://github.com/rust-embedded/cortex-m-rt/commit/d5f2a94d23e921a26350f9d7ba79af3cf8e6bbab#diff-21b67a00d2a7d05c7aa7f0fe1a2f8adfR96

From the tests I just ran, removing this preserves rust-lld and arm-none-eabi-ld linkability, but fixes arm-none-eabi-gcc, so perhaps we should revert that change (and backport to 0.6.x)?

Possibly we should update the CI tests to run against arm-none-eabi-gcc as well as the current arm-none-eabi-ld and rust-lld.

adamgreig:


adamgreig:
Looks like adding arm-none-eabi-gcc to CI has revealed another issue with floating point targets.

adamgreig:
I've been investigating a bit further and actually the `-nostartfiles` to gcc _also_ fixes this problem, without changing the linker script. However, it still looks to me like we don't need to put `__etext` where we do; it makes no difference to the resulting binaries with any of the linkers, so possibly we should remove it anyway.

I definitely think we should still add gcc to the CI scripts.

@therealprof any recollection what the purpose of that line was in #207?

adamgreig:
@therealprof mentioned that the bug was hit in rust-lld when _stext was offset: https://github.com/rust-lang/rust/issues/65391

I'll do some more tests in that configuration. Also, we should probably add a MSRV run to the CI, perhaps 1.32.0.

adamgreig:
Testing with a memory.x-overridden stext shows we do still need the full modification to the linker script. I'm going to close this PR and open a new one just for adding arm-none-eabi-gcc to the CI. Sorry for the noise!

PR #262, opened 2020-04-04
Add testing linker=arm-none-eabi-gcc and MSRV to CI

adamgreig:
Replacing #260.

This PR extends our current tests with `linker=rust-lld` and `linker=arm-none-eabi-ld` to include `linker=arm-none-eabi-gcc`, since those options are all included in our example [`.cargo/config`](https://github.com/rust-embedded/cortex-m-rt/blob/master/.cargo/config). It looks like another linker only adds a handful of seconds to CI, since most time is spent building dependencies once.

It also adds a test with Rust 1.32.0 to the CI as a candidate MSRV. Building with 1.31.0 fails because of the dev-dependency on cortex-m-semihosting 0.3.5:

```
error[E0658]: using the `?` macro Kleene operator for "at most one" repetition is unstable (see issue #48075)
   --> /home/adam/.cargo/registry/src/github.com-1ecc6299db9ec823/cortex-m-semihosting-0.3.5/src/macros.rs:111:25
    |
111 |     ($($val:expr),+ $(,)?) => {
    |                         ^

error: expected `*` or `+`
```
That's just a test error which end users wouldn't experience, so we could consider making 1.31.0 the MSRV and working around the c-m-semihosting issue. I don't know if there are other 1.31.0 issues.

Finally the PR removes the travis check preventing builds on pushes to master. In principle we know that builds to master succeed because bors tests them before pushing, so adding the extra check mostly means we get a well-defined build status from Travis for flags etc. The build times for cortex-m-rt (around 3min overall) are a bit longer than r0 (where we did the same thing), so we could just run a minimal test here instead. I don't think it's a significant overhead given how infrequently we push to master, though.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,53 +4,62 @@ matrix:
   allow_failures:
     - rust: nightly
   include:
+    # MSRV 1.32.0 #############################################################
+    - env: TARGET=x86_64-unknown-linux-gnu
+      rust: 1.32.0
+
+    - env: TARGET=thumbv6m-none-eabi
+      rust: 1.32.0
+
+    - env: TARGET=thumbv7m-none-eabi
+      rust: 1.32.0
+
+    - env: TARGET=thumbv7em-none-eabi
+      rust: 1.32.0
+
+    - env: TARGET=thumbv7em-none-eabihf
+      rust: 1.32.0
+
+    - env: TARGET=thumbv8m.main-none-eabi
+      rust: 1.32.0
+
+    # Stable ##################################################################
     - env: TARGET=x86_64-unknown-linux-gnu
       rust: stable
-      if: (branch = staging OR branch = trying) OR (type = pull_request AND branch = master)
 
     - env: TARGET=thumbv6m-none-eabi
       rust: stable
-      if: (branch = staging OR branch = trying) OR (type = pull_request AND branch = master)
 
     - env: TARGET=thumbv7m-none-eabi
       rust: stable
-      if: (branch = staging OR branch = trying) OR (type = pull_request AND branch = master)
 
     - env: TARGET=thumbv7em-none-eabi
       rust: stable
-      if: (branch = staging OR branch = trying) OR (type = pull_request AND branch = master)
 
     - env: TARGET=thumbv7em-none-eabihf
       rust: stable
-      if: (branch = staging OR branch = trying) OR (type = pull_request AND branch = master)
 
     - env: TARGET=thumbv8m.main-none-eabi
       rust: stable
-      if: (branch = staging OR branch = trying) OR (type = pull_request AND branch = master)
 
+    # Nightly #################################################################
     - env: TARGET=x86_64-unknown-linux-gnu
       rust: nightly
-      if: (branch = staging OR branch = trying) OR (type = pull_request AND branch = master)
 
     - env: TARGET=thumbv6m-none-eabi
       rust: nightly
-      if: (branch = staging OR branch = trying) OR (type = pull_request AND branch = master)
 
     - env: TARGET=thumbv7m-none-eabi
       rust: nightly
-      if: (branch = staging OR branch = trying) OR (type = pull_request AND branch = master)
 
     - env: TARGET=thumbv7em-none-eabi
       rust: nightly
-      if: (branch = staging OR branch = trying) OR (type = pull_request AND branch = master)
 
     - env: TARGET=thumbv7em-none-eabihf
       rust: nightly
-      if: (branch = staging OR branch = trying) OR (type = pull_request AND branch = master)
 
     - env: TARGET=thumbv8m.main-none-eabi
       rust: nightly
-      if: (branch = staging OR branch = trying) OR (type = pull_request AND branch = master)
 
 before_install: set -e
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,24 +4,24 @@ matrix:
   allow_failures:
     - rust: nightly
   include:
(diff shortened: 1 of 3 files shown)


adamgreig:
Unfortunately it looks like this is failing on 1.32.0 on travis, despite succeeding locally.

adamgreig:
I guess this just highlights the use of having MSRV be tested in CI! It looks like `cargo test --doc` fails because it pulls in compiletest_rs which pulls in rustfix which pulls in anyhow which doesn't work on 1.32.0. I only ran the build tests and not `cargo test`, so didn't catch it.

adamgreig:
Our current minimum Rust version that can execute `ci/script.sh` without error is **1.41.0**, mainly because of [this doctest](https://github.com/rust-embedded/cortex-m-rt/blob/master/src/lib.rs#L375). With Rust 1.40.0, we get:

```
---- src/lib.rs -  (line 374) stdout ----
error[E0433]: failed to resolve: maybe a missing crate `core`?
 --> src/lib.rs:375:5
  |
3 | use core::mem::MaybeUninit;
  |     ^^^^ maybe a missing crate `core`?
```

Changing `core` to `std` in that doctest lets us run down to Rust 1.39.0, which I think is the first time proc macros could be applied to functions arguments? At 1.38.0, we get:

```
---- src/lib.rs - exception (line 638) stdout ----
error[E0658]: attributes on function parameters are unstable
 --> src/lib.rs:641:1
  |
5 | #[exception]
  | ^^^^^^^^^^^^
  |
  = note: for more information, see https://github.com/rust-lang/rust/issues/60406
```
I don't understand what changed in 1.41.0 that allows doctests to use `core` without error, but if we could work around that (or disable this doctest) we could suport MSRV of 1.39.0. I'm not totally clear on the E0658 but presumably that's what stops us going any lower.

adamgreig:
I've fixed the doc test by adding `# extern crate core;` which seems to work on both 1.39.0 and stable. I updated the reported MSRV and Travis test to also use 1.39.0, so hopefully this build will pass.

I don't know if we want to try and target a lower MSRV on this quite foundational crate.

therealprof:
Yes, I do think we need to target a lower MSRV.

adamgreig:
As stands, we can't run our tests on 1.32 because of compiletest_rs depending on rustfix depending on anyhow, and I think if we can't run tests we probably can't claim a MSRV.

By 1.38 we can't run because #205 adds attributes to the parameters of the generated functions for entry/exception, which aren't stabilised until 1.39. I think we'd need to do at least a non trivial amount of re-engineering to bring the MSRV down to 1.38 or lower.

adamgreig:
Is it worth merging this PR now to at least get an MSRV test in the CI, and add gcc to linker testing, and then later we can do another PR to bring down the (now tested) MSRV?

adamgreig:
With the following diff (removing `#[allow(non_snake_case)]` from the generated handlers' attributes) we can build on 1.38.0, but 1.37.0 fails because our dev-dependency on `tester` uses `std::time::Duration::as_secs_f64()` which wasn't stabilised until 1.38.

```diff
diff --git a/macros/src/lib.rs b/macros/src/lib.rs
index 09881b85..8e15446c 100644
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -66,7 +66,7 @@ pub fn entry(args: TokenStream, input: TokenStream) -> TokenStream {
         // Note that we use an explicit `'static` lifetime for the entry point arguments. This makes
         // it more flexible, and is sound here, since the entry will not be called again, ever.
         syn::parse::<FnArg>(
-            quote!(#[allow(non_snake_case)] #(#attrs)* #ident: &'static mut #ty).into(),
+            quote!(#(#attrs)* #ident: &'static mut #ty).into(),
         )
         .unwrap()
     }));
@@ -297,7 +297,7 @@ pub fn exception(args: TokenStream, input: TokenStream) -> TokenStream {
                 let ty = &statik.ty;
                 let attrs = &statik.attrs;
                 syn::parse::<FnArg>(
-                    quote!(#[allow(non_snake_case)] #(#attrs)* #ident: &mut #ty).into(),
+                    quote!(#(#attrs)* #ident: &mut #ty).into(),
                 )
                 .unwrap()
             }));
@@ -403,7 +403,7 @@ pub fn interrupt(args: TokenStream, input: TokenStream) -> TokenStream {
         let ident = &statik.ident;
         let ty = &statik.ty;
         let attrs = &statik.attrs;
-        syn::parse::<FnArg>(quote!(#[allow(non_snake_case)] #(#attrs)* #ident: &mut #ty).into())
+        syn::parse::<FnArg>(quote!(#(#attrs)* #ident: &mut #ty).into())
             .unwrap()
     }));
     f.block.stmts = iter::once(
```

jonas-schievink:
Hmm, do we just slap the attributes on the `static mut` onto the argument and parameter? That doesn't sound right.

adamgreig:
In our currently published 0.6.12, this is the expansion:

```rust
#[interrupt]
fn TIM2() {
    static mut COUNT: i32 = 0;
}
```

```rust
#[doc(hidden)]
#[export_name = "TIM2"]
pub unsafe extern "C" fn __cortex_m_rt_TIM2_trampoline() {
    __cortex_m_rt_TIM2({
        static mut COUNT: i32 = 0;
        &mut COUNT
    })
}
fn __cortex_m_rt_TIM2(#[allow(non_snake_case)] COUNT: &mut i32) {
    {
        extern crate cortex_m_rt;
        interrupt::TIM2;
    }
}

korken89:
bors retry

PR #270, opened 2020-05-19
Add GitHub actions for CI

therealprof:
This changes CI from Travis CI to GHA which should not only vastly simplify the maintenance (by getting rid of pulling custom binaries from japaric's repository) but also speeds up the CI process and adds rustfmt as well as clippy checks for good measure.

There's still room for optimisation but this has taken me long enough and plenty of sweat to get to this stage so my recommendation would be to merge it.


therealprof:
bors try

therealprof:
bors try

adamgreig:
LGTM. Only comment is whether we want bors to require the rustfmt check passes before merging?

therealprof:
> LGTM. Only comment is whether we want bors to require the rustfmt check passes before merging?

Can do. As you probably noticed I already ran into problems with bad formatting and failing builds. ;)

PR #271, opened 2020-05-19
Get rid of #[deny(warnings)]

therealprof:
This is considered harmful as the warnings may change with any compiler
release.

### DIFF
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -2,8 +2,6 @@
 //!
 //! Do not use this crate directly.
 
-#![deny(warnings)]
-
 extern crate proc_macro;
 
 use proc_macro::TokenStream;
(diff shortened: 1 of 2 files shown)


therealprof:
r? @jonas-schievink 

This is blocking the attempt to use clippy via GHA cf. #270

jonas-schievink:
We should still keep at least a CI job that sets `RUSTFLAGS=-Dwarnings` to make sure no code with warnings is checked in. Other than that this looks good!

adamgreig:
Indeed, I thought the general idea was to not _release_ code with deny(warnings) as it breaks downstream, but still to run all the CI jobs with `-Dwarnings` to keep our repo code warning-free.

therealprof:
I'm working on GHA setup and part of this will be clippy so we will get nice annotations...

jonas-schievink:
The clippy job can leave `-Dwarnings` off, but denying the built-in lints is usually a good idea (at least in CI)

therealprof:
We can certainly do that, but probably should only add them to MSRV builds.

PR #276, opened 2020-07-03
Replace __ONCE__ with Cargo links key

### LINKED ISSUE #275: "error: cortex-m-rt appears more than once in the dependency graph"
I've been reading some of the emitted ASM of my application, and I've found that "error: cortex-m-rt appears more than once in the dependency graph" is present in the generated binary (-O3) and is referenced 126 times (63 loads I think) in my assembly. What's up with that, if the error is only there for compile-time?

adamgreig:
This would fix #275. We use the links key in cortex-m already [here](https://github.com/rust-embedded/cortex-m/blob/master/Cargo.toml#L16). See also https://github.com/rust-embedded/wg/issues/467.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ readme = "README.md"
 repository = "https://github.com/rust-embedded/cortex-m-rt"
 version = "0.6.11"
 autoexamples = true
+links = "cortex-m-rt" # Prevent multiple versions of cortex-m-rt being linked
 
 [dependencies]
 r0 = "1.0"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -689,6 +689,14 @@ pub use macros::exception;
 /// [rfc1414]: https://github.com/rust-lang/rfcs/blob/master/text/1414-rvalue_static_promotion.md
 pub use macros::pre_init;
 
+// We export this static with an informative name so that if an application attempts to link
+// two copies of cortex-m-rt together, linking will fail. We also declare a links key in
+// Cargo.toml which is the more modern way to solve the same problem, but we have to keep
+// __ONCE__ around to prevent linking with versions before the links key was added.
+#[export_name = "error: cortex-m-rt appears more than once in the dependency graph"]
+#[doc(hidden)]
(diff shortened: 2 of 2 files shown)


jonas-schievink:
:+1: to adding `links`, all rt and peripheral crates should have that. When I added it to cortex-m though I left the static in place, since that's still needed to prevent linking with older versions. If we don't do that we effectively lose the duplicate detection between newer and older cmrt versions.

adamgreig:
In cortex-m, the static you're talking about is `CORE_PERIPHERALS: bool`, right? Don't we need that static irregardless of linkage, so at most we could have removed the no_mangle?

The issue here is that because the static is zero sized, disassemblers mix it up with whatever other statics end up at the same address.

Nevertheless you're totally right and removing the static in the same release as adding the links key does stop duplicate detection between versions which is a pain. Any other ideas for fixing the debug issue? Making the static non-zero-sized seems undesirable.

jonas-schievink:
Ah, I see, yeah. That static has another purpose and need to be kept. I don't immediately see how to fix this debugging issue other than fixing the debug tooling (which arguably is the "proper" fix).

adamgreig:
@jonas-schievink I've updated to keep __ONCE__ (though added a comment about this) and still add the links key, which I guess i the best we can do for now.

therealprof:
I guess we should release this ASAP then?

adamgreig:
I don't think we'll want to remove __ONCE__ for at least one or two minor/major releases, so there's probably no specific rush, but we probably do want a new c-m-rt patch release soonish.

therealprof:
Yes, a patch release to get "links" out into the field.

adamgreig:
Sure, it's worth doing eventually, but there's no rush since it won't make any difference until there's two releases using it (which could then be detected attempting to link together).

therealprof:
Maybe I'm a bit over-cautious  but I'm wary about any new problems which could occur with the `links` approach in our context and just haven't been detected yet. AFAIK `links` has only/mostly been used for sys crates so far.

PR #279, opened 2020-07-12
Add barriers after FPU enabling

thalesfragoso:
This only seems to be required for M7 cores, but since we can't know the specific core, we can't filter on that.

I thought about using the SCB's `RegisterBlock` from cortex-m to enable the FPU, but the [registers](https://docs.rs/cortex-m/0.6.2/cortex_m/peripheral/scb/struct.RegisterBlock.html) are just the generics `RW` which isn't that different from what we're doing right now, and we would need `Peripherals::steal` to use the methods from SCB which would set the `CORE_PERIPHERALS`/`TAKEN` flag.

And one question, with this change, can we get rid of the `trampoline()` function? I don't think the compiler will do reordering across foreign functions calls (dsb/isb), but I might be missing something.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,9 +19,9 @@ links = "cortex-m-rt" # Prevent multiple versions of cortex-m-rt being linked
 [dependencies]
 r0 = "1.0"
 cortex-m-rt-macros = { path = "macros", version = "=0.6.11" }
+cortex-m = "0.6"
 
 [dev-dependencies]
-cortex-m = "0.6"
 panic-halt = "0.2.0"
 cortex-m-semihosting = "0.3"
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -916,12 +916,12 @@ pub unsafe extern "C" fn Reset() -> ! {
     r0::zero_bss(&mut __sbss, &mut __ebss);
     r0::init_data(&mut __sdata, &mut __edata, &__sidata);
 
+    #[allow(clippy::match_single_binding)]
     match () {
         #[cfg(not(has_fpu))]
         () => main(),
         #[cfg(has_fpu)]
         () => {
-            // We redefine these here to avoid pulling the `cortex-m` crate as a dependency
             const SCB_CPACR: *mut u32 = 0xE000_ED88 as *mut u32;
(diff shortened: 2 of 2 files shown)


adamgreig:
> And one question, with this change, can we get rid of the trampoline() function? I don't think the compiler will do reordering across foreign functions calls (dsb/isb), but I might be missing something.

The problem isn't reordering but any use of VFP functions. For example [in this snippet](https://rust.godbolt.org/z/cfhaWY) because `foo` takes an f32 but first calls a function, it pushes the input onto the stack before calling the function, and `vpush` and `vmov` both fault if the FPU is disabled. Of course in our case `main` won't take a float argument; off the top of my head I can't make up an example that fails in our case but I don't think there's any reason to believe it's impossible. I'd be happy if someone could demonstrate otherwise of course.

adamgreig:
On this actual PR: I agree we should have the fences; do you think there's any downsides to bringing in a dependency on cortex-m though?

> I thought about using the SCB's RegisterBlock from cortex-m to enable the FPU, but the registers are just the generics RW which isn't that different from what we're doing right now, and we would need Peripherals::steal to use the methods from SCB which would set the CORE_PERIPHERALS/TAKEN flag.

I think we could just use the unsafe `ptr()` methods, which would be neater than redefining these memory addresses if we are including cortex-m.

thalesfragoso:
> I think we could just use the unsafe ptr() methods, which would be neater than redefining these memory addresses if we are including cortex-m.

The problem is that `ptr()` gives us a `RegisterBlock` not the zero sized `SCB` type that has the helper methods. Also, differently from the usual pacs, all of the registers in `scb::RegisterBlock` are just generics `RW` without any field's methods (maybe we should change that ?). Using `cortex-m` would only get rid of

```rust
const SCB_CPACR: *mut u32 = 0xE000_ED88 as *mut u32;
```
we would still need the others constants and all the bit shifting that we are doing now. Because of that I thought it wasn't really worth the code change and the (small) increase of optimization "dependency".

adamgreig:
> The problem is that ptr() gives us a RegisterBlock

Oh, yea, that's annoying. All issues to be addressed by a hypothetical "cortex-m-pac"/"cortex-m-hal" split I guess. Seems fine to leave it for now then.

bors merge

PR #280, opened 2020-07-22
Make Vector public in interrupt example

Tiwalun:
If the `Vector`union is not public, the example code leads to a compilation error because `pub static __INTERRUPTS` leaks the private type `Vector`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -283,7 +283,7 @@
 //! IRQ number = 2, and `Bar`, with IRQ number = 4.
 //!
 //! ```no_run
-//! union Vector {
+//! pub union Vector {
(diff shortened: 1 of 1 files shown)


jonas-schievink:
Hmm, why is CI not catching this? It does build doctests.

jonas-schievink:
Oh, because rustdoc puts everything in an `fn main()`, and `pub` does nothing there.

Tiwalun:
I've tried adding an explicit `fn main()`, but that doesn't help, the test still passes without the `pub`.

adamgreig (approved):
Tests aside let's at least get this merged in, thanks @Tiwalun!

bors merge

Issue #281, opened 2020-07-24
How to utilize CCM SRAM for uninit statics?


tstellanova:
The documentation in `lib.rs` mentions using CCRAM for `_stack_start` but it does not seem possible to use CCM for eg uninit statics because there is no section in [link.x.in](https://github.com/rust-embedded/cortex-m-rt/blob/master/link.x.in#L60) that directs its output to CCM.  

I realize not all cortex-m devices have something like CCM, but having a placeholder section available in `link.x` for that, when it is available, would allow developers to take advantage of it when it is available.

adamgreig:
You can do this without any changes to cortex-m-rt, though perhaps it could be better documented.

In your `memory.x`, write something like:

```
MEMORY
{
    FLASH  (rx) : ORIGIN = 0x08000000, LENGTH = 1024K
    RAM    (rw) : ORIGIN = 0x20000000, LENGTH = 128K
    CCMRAM (rw) : ORIGIN = 0x10000000, LENGTH = 64K
}

SECTIONS
{
    .ccmram (NOLOAD) : ALIGN(4)
    {
        *(.ccmram .ccmram.*);
        . = ALIGN(4);
    } > CCMRAM
} INSERT AFTER .bss;

/* Place stack in CCMRAM (optional) */
_stack_start = ORIGIN(CCMRAM) + LENGTH(CCMRAM);
```

You can then place uninit statics into CCMRAM:

```rust
#[link_section=".ccmram"]
static mut BUF: [u8; 1024] = [0u8; 1024];
```

tstellanova:
Thanks for the assistance-- I did not realize SECTIONS could be modified this way.  This works!

therealprof:
Is this documented somewhere in this detail? If not it probably should be.

tstellanova:
> Is this documented somewhere in this detail? If not it probably should be.

I'd be happy to write this up somewhere.  Where's the right spot?  Maybe the lib [comments](https://github.com/rust-embedded/cortex-m-rt/blob/98e4c3c8597f1dfa52d55616c7ce7cf90682d95b/src/lib.rs#L52)

PR #282, opened 2020-07-31
Doc adding memory sections in memory.x

tstellanova:
As noted in #281 it is possible to define additional memory sections in `memory.x`.
Added some documentation on how to do this.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -392,8 +392,10 @@
 //! Some microcontrollers provide additional memory regions beyond RAM and FLASH.
 //! For example, some STM32 devices provide "CCM" or core-coupled RAM that is
 //! only accessible from the core. In order to access these using
-//! `#[link_section=".foo"]` from your code, you need to modify `memory.x`
+//! [`link_section`] attributes from your code, you need to modify `memory.x`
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -387,6 +388,39 @@
 //! undefined behavior. At some point in the future we may add an attribute to safely place static
 //! variables in this section.
 //!
+//! ## Extra Sections
+//!
+//! Some microcontrollers provide additional memory regions beyond RAM and FLASH.
+//! For example, some STM32 devices provide "CCM" or core-coupled RAM that is
+//! only accessible from the core. In order to access these using
+//! `#[link_section=".foo"]` from your code, you need to modify `memory.x`
```
adamgreig, line 395: Maybe you could either use the same link section as below (i.e. `#[link_section=".ccmram"]`) to keep the example clearer, or just write something like "In order to access these using the link_section attribute in your code, ...", perhaps with a link to https://doc.rust-lang.org/reference/abi.html#the-link_section-attribute.

tstellanova: Good suggestions-- I've updated.

PR #285, opened 2020-08-04
Try combining some steps to speed up CI

therealprof:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,9 +13,6 @@ jobs:
         # All generated code should be running on stable now
         rust: [nightly, stable, 1.39.0]
 
-        # The default target we're compiling on and for
-        TARGET: [x86_64-unknown-linux-gnu, thumbv6m-none-eabi, thumbv7m-none-eabi, thumbv7em-none-eabi, thumbv7em-none-eabihf, thumbv8m.main-none-eabi]
-
         include:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -27,10 +24,20 @@ jobs:
         with:
           profile: minimal
           toolchain: ${{ matrix.rust }}
-          target: ${{ matrix.TARGET }}
+          target: thumbv6m-none-eabi thumbv7m-none-eabi thumbv7em-none-eabi thumbv7em-none-eabihf thumbv8m.main-none-eabi
```
jonas-schievink, line 27: I don't think this action supports lists https://github.com/actions-rs/toolchain/pull/62

jonas-schievink: It might be easiest to just invoke rustup manually, it *can* take a list of targets to install

therealprof: Yes indeed. I'm alreay working on it.

therealprof:
Hm, no benefit here wall clock wise. But we're using way less builders which might be benefitial anyway.

jonas-schievink:
We could also collapse the Clippy and rustfmt builders into one

therealprof:
Yes we could but I actually prefer them to be separate because that makes it easier for me to add them to other crates. 😅

Issue #188, opened 2019-04-12
_stext fails to move .text with lld


yodaldevoid:
I have an older program/library that I am moving to lld so that I can work with ease on more than just Linux. The code in question is https://github.com/yodaldevoid/mkl26/tree/lld_port. Please note that the example/library currently require nightly and should be built for thumbv6m-none-eabi.

To start, the code compiles and runs correctly using gcc's linker.

When attempting to compile the `blink` example with lld using `cargo build --example blink` I get the following error:
```
error: linking with `rust-lld` failed: exit code: 1
  |
  = note: "rust-lld" "-flavor" "gnu" "-L" "C:\\Users\\ga29s\\.rustup\\toolchains\\nightly-x86_64-pc-windows-msvc\\lib\\rustlib\\thumbv6m-none-eabi\\lib" "E:\\mkl26\\target\\thumbv6m-none-eabi\\debug\\examples\\blink-288c8e46abd20a15.blink.2hpe9hsl-cgu.0.rcgu.o" "-o" "E:\\mkl26\\target\\thumbv6m-none-eabi\\debug\\examples\\blink-288c8e46abd20a15" "--gc-sections" "-L" "E:\\mkl26\\target\\thumbv6m-none-eabi\\debug\\deps" "-L" "E:\\mkl26\\target\\debug\\deps" "-L" "E:\\mkl26\\target\\thumbv6m-none-eabi\\debug\\build\\mkl26-36049ce57ff99828\\out" "-L" "E:\\mkl26\\target\\thumbv6m-none-eabi\\debug\\build\\mkl26-36049ce57ff99828\\out" "-L" "E:\\mkl26\\target\\thumbv6m-none-eabi\\debug\\build\\cortex-m-rt-6e1f169148af1ece\\out" "-L" "C:\\Users\\ga29s\\.rustup\\toolchains\\nightly-x86_64-pc-windows-msvc\\lib\\rustlib\\thumbv6m-none-eabi\\lib" "-Bstatic" "C:\\Users\\ga29s\\AppData\\Local\\Temp\\rustcVhEipN\\libcortex_m_rt-193fc16ad349c2cc.rlib" "C:\\Users\\ga29s\\.rustup\\toolchains\\nightly-x86_64-pc-windows-msvc\\lib\\rustlib\\thumbv6m-none-eabi\\lib\\libcompiler_builtins-41a0f49e919d94c0.rlib" "-Tlink.x" "-Bdynamic"
[...]

PR #286, opened 2020-08-07
Fix possible overflow of .data region in FLASH

Guilucand:
Hi,

i found a bug on the size checking of the FLASH region, that in some cases allows it to be larger than it's defined size.
This happens because the .data section is placed on a FLASH address (at the end of __erodata) but it's not checked against the FLASH region size.
Changing the section as shown solves the issue.

### DIFF
--- a/link.x.in
+++ b/link.x.in
@@ -106,14 +106,14 @@ SECTIONS
 
   /* ## Sections in RAM */
   /* ### .data */
-  .data : AT(__erodata) ALIGN(4)
+  .data : ALIGN(4)
   {
     . = ALIGN(4);
     __sdata = .;
     *(.data .data.*);
     . = ALIGN(4); /* 4-byte align the end (VMA) of this section */
     __edata = .;
-  } > RAM
+  } > RAM AT>FLASH
 
   /* LMA of .data */
   __sidata = LOADADDR(.data);


jonas-schievink:
Could you maybe provide an example where this happens? I'd expect the linker to catch an issue like that, even without these changes.

Guilucand:
Hi,
i've made an example repository that triggers the unwanted behavior:
[https://github.com/Guilucand/cortex-m-rt-flash-overflow-example]()

The problem only happens when the flash size is smaller than the ram size, 
so i set the flash size to 64K and the ram size to 256K, and allocate an array of 100K bytes.
If you build the example with:
`cargo build --release`
the linker throws an error because the array is in .rodata and it's checked against the FLASH bounds.
If instead you build the example with:
`cargo build --release --features trigger-overflow`
the linker produces a binary with a .data section size of 100K, overflowing the FLASH region.

adamgreig:
It looks like this PR basically half-reverts https://github.com/rust-embedded/cortex-m-rt/commit/d5f2a94d23e921a26350f9d7ba79af3cf8e6bbab since we used to use `AT > FLASH` and then changed to `AT(__erodata)` instead, to fix #188 and https://github.com/rust-lang/rust/issues/65391.

Maybe @therealprof will recall why that was necessary, and perhaps we should check we haven't re-introduced those bugs.

therealprof:
I guess the thinking was that the `.erodata` would be sufficient to constrain the placement of the sections which is why I've removed the `AT > FLASH`. It is quite possible that this assumption doesn't hold under some conditions. I have no objection to readding this but maybe it would be possible to add regression test case for this.

jonas-schievink:
Since that PR has fixed linkage errors, but linking still works, I guess it didn't reintroduce those problems

PR #291, opened 2020-08-29
Avoid depending on `cortex-m`

jonas-schievink:
I'd like to use `cortex-m-rt` to write test executables for `cortex-m`. This would pull in multiple versions of `cortex-m` when c-m-rt depends on it, which does not work.

Replace the Rust FPU trampoline with one written in asm to get rid of the cortex-m dependency.


adamgreig (approved):
LGTM

bors merge

PR #292, opened 2020-09-02
[ARMv6-M] initialize the LR register

japaric:
the ARMv6-M Architecture Reference Manual (ARM DDI 0419D) indicates in section
B1.5.5 "Reset behavior" that the LR (Link Register) starts in an unknown state
when the Reset handler is taken and that its "Value must be initialised by
software"

So this PR does that: it initializes the LR register to 0xFFFF_FFFF (-1) first
thing in the Reset handler (only for v6). The manual doesn't say which value to
use so I decided to use the value used by the ARMv7-M (v7 sets LR to 0xFFFF_FFFF
before invoking the Reset handler; see its Architecture Manual for details).

The values of LR (these are pushed onto the stack in function preludes) are used
to unwind the stack (e.g. GDB's `backtrace` or a future `cortex_m_panic_unwind`
handler). Having the initial stack frame use a known value on all Cortex-M
variants makes it easier to implement `panic_unwind` and avoids virtual
unwinders like GDB `backtrace` trying to unwind beyond the `Reset` handler

Note that this implementation uses a trampoline that runs before `Reset` to set
LR on v6. This is required because the prelude of the `Reset` routine will push
LR onto the stack; we want that LR value to be -1. Calling `register::lr::write`
from `Reset` would perform the write after LR has been pushed onto the stack and
that's too late

*NOTE* this is a PR against v0.6.x. I'm using that branch as 0.6.x is the latest minor release. I can open a PR against master after this one has been approved.


jonas-schievink (approved):
LGTM. I'd like to land this in master first though, to avoid having something in v0.6.x that isn't in master.

japaric:
PR against the main branch: #293

therealprof (approved):
LGTM

therealprof:
Gah!

`+curl -L 'https://developer.arm.com/-/media/Files/downloads/gnu-rm/7-2018q2/gcc-arm-none-eabi-7-2018-q2-update-linux.tar.bz2?revision=bc2c96c0-14b5-4bb4-9f18-bceb4050fee7?product=GNU%20Arm%20Embedded%20Toolchain,64-bit,,Linux,7-2018-q2-update'`

therealprof:
Hm, genuine error:
```
error: tests/compile-fail/interrupt-invalid.rs:22: unexpected error: '22:4: 22:7: no variant or associated item 
...
command: "rustc" "tests/compile-fail/interrupt-invalid.rs" "-L" "/tmp" "--target=x86_64-unknown-linux-gnu" "--error-format" "json" "-C" "prefer-dynamic" "-o" "/tmp/interrupt-invalid.stage-id" "-L" "target/debug" "-L" "target/debug/deps" "-C" "panic=abort" "-L" "/tmp/interrupt-invalid.stage-id.aux" "-A" "unused"
```

No idea why though.

jonas-schievink:
rustc error changed, and we don't run CI on PRs to the v0.6.x branch so we didn't see the failure earlier

jonas-schievink (approved):
bors r=therealprof

therealprof:
Huh? Why didn't that dismiss the previous review. Will check the configuration.

PR #293, opened 2020-09-02
[ARMv6-M] initialize the LR register

japaric:
NOTE: same as PR #292

the ARMv6-M Architecture Reference Manual (ARM DDI 0419D) indicates in section
B1.5.5 "Reset behavior" that the LR (Link Register) starts in an unknown state
when the Reset handler is taken and that its "Value must be initialised by
software"

So this PR does that: it initializes the LR register to 0xFFFF_FFFF (-1) first
thing in the Reset handler (only for v6). The manual doesn't say which value to
use so I decided to use the value used by the ARMv7-M (v7 sets LR to 0xFFFF_FFFF
before invoking the Reset handler; see its Architecture Manual for details).

The values of LR (these are pushed onto the stack in function preludes) are used
to unwind the stack (e.g. GDB's `backtrace` or a future `cortex_m_panic_unwind`
handler). Having the initial stack frame use a known value on all Cortex-M
variants makes it easier to implement `panic_unwind` and avoids virtual
unwinders like GDB `backtrace` trying to unwind beyond the `Reset` handler

Note that this implementation uses a trampoline that runs before `Reset` to set
LR on v6. This is required because the prelude of the `Reset` routine will push
LR onto the stack; we want that LR value to be -1. Calling `register::lr::write`
from `Reset` would perform the write after LR has been pushed onto the stack and
that's too late


therealprof (approved):
LGTM, thanks!

bors r+

PR #294, opened 2020-09-04
add CFI and size info

japaric:
like it was done in rust-embedded/cortex-m#216 and rust-embedded/cortex-m#212


jonas-schievink (approved):
bors r+

PR #295, opened 2020-09-04
[backport] add CFI and size info to external assembly

japaric:
backport of #294


therealprof (changes requested):
LGTM but we should have a CHANGELOG entry since we're probably going to publish this, right?

jonas-schievink (approved):
bors r+

therealprof (approved):
LGTM. Thanks!

bors r+

PR #301, opened 2020-11-25
Initialize RAM in assembly

### LINKED ISSUE #300: RAM initialization code violates pointer provenance
https://github.com/rust-embedded/cortex-m-rt/blob/96525a64197049d11cfc8cb5cc2c4dc9b5240e42/src/lib.rs#L939-L959

This code uses a pointer to a single `u32`, but writes an arbitrary amount of data beyond the `u32`. This is UB, as it violates the contract of `ptr::offset`, which is called by r0. Namely:

> Both the starting and resulting pointer must be either in bounds or one byte past the end of the same allocated object. Note that in Rust, every (stack-allocated) variable is considered a separate allocated object.

jonas-schievink:
Fixes #300


asm.s
```diff
@@ -62,6 +62,38 @@ PreResetTrampoline:
...
+  ldr r0,=__sdata
+  # ...up to here
+  ldr r1,=__edata
+  # copy from here
+  ldr r2,=__sidata
+2:
+  cmp r1, r0
+  beq 3f
+  # load 1 word from r2 to r3, inc r2
+  ldm r2!, {r3}
```
nagisa, line 90: Why not utilize more registers here?

therealprof: Why bother?

nagisa: Well… I haven't done any measurements, but intuition says it may be faster – as in, have less trouble saturating available memory bandwidth.

I tried to phrase this as a mild suggestion ^^

jonas-schievink: Yeah we can certainly optimize this for more speed if we want, but the existing impl is optimized for code size.

I'm not sure how we can utilize MPU requirements here though, this code runs before the MPU can be enabled (and some chips we target don't have an MPU).

therealprof: I think simplicity is key here. At this point we're running with default clocks anyway, so it's not going to be terribly fast whatever we might do.

nagisa: The reason why MPU came up here is because it _might_ be enabled here (by e.g. a bootloader that could've been run before this), and MPU is the only thing AFAIK that would prevent code copying algorithm that goes OOB invalid.

As another point towards validity of using multiple registers, regardless of OOB reads/writes, IIRC IAR's initialization code also uses multiple registers for their thumb initialization code. Though they may have more stringent alignment requirements for their sections too.

> the existing impl is optimized for code size

Fair point. There isn't any additional cost in specifying more registers to `stm` and `ldm` instructions, but I think you'd still need some code to align the pointers a in a very specific way before the loop.

pftbest: > some code to align the pointers a in a very specific way before the loop.

The pointers can be aligned manually in the linker script, so no modification to the code is needed.

adamgreig:
GitHub won't let me comment outside the diff itself, so..

* The comments above `PreResetTrampoline` still refer to it being an ARMv6-M-only thing that sets LR; they need updating to say it's actually the reset vector on all architectures and initialises bss and data.
* This change causes us to write LR on every architecture, instead of only ARMv6-M, unnecessarily. Not sure what the cleanest way to only do it on v6 is, but I'd rather we didn't always set it on v7.
* Not sure if the names are still right. Perhaps the assembly stub should be `Reset` now, since it's always the reset vector and contains memory initialisation?
* We still expose `pre_init` before bss/data are initialised, which I think is the most useful option but the whole reason for this PR is because "life before main" is UB; we should update the documentation around `pre_init` in this PR too. Or add `post_init`?

jonas-schievink:
> This change causes us to write LR on every architecture, instead of only ARMv6-M, unnecessarily. Not sure what the cleanest way to only do it on v6 is, but I'd rather we didn't always set it on v7.

It actually gets immediately overwritten by the call to `__pre_init`, so it seems like this might break unwinding? Seems like we either have to write it again after `__pre_init` returns or save it in another register and write the appropriate CFI directive (no idea what that would look like).

jonas-schievink:
Updated the docs in the assembly. Also made it restore LR after __pre_init returns.

> We still expose `pre_init` before bss/data are initialised, which I think is the most useful option but the whole reason for this PR is because "life before main" is UB; we should update the documentation around `pre_init` in this PR too. Or add `post_init`?

Do we want to remove/deprecate `#[pre_init]` entirely? It already has a big warning about accessing statics inside of it. If we do want to remove or deprecate it I think we might want to wait until `asm!`/`#[naked]` are stable so that users have a viable alternative by writing their `__pre_init` in inline assembly.

`post_init` is just `entry`, so I don't see what that would let people do that they can't already.

Sympatron:
I think #303 shows that `#[pre_init]` can be useful in some cases. Without it there would be no way to do this. With enough warnings in the docs, I would vote to leave it there in case someone needs it.

jonas-schievink:
To be clear, we're not taking away the ability to run code before RAM initialization *in general*, but allowing arbitrary Rust code in a `#[pre_init]` function is problematic.

adamgreig (approved):
Thanks!

bors merge

Issue #145, opened 2018-10-27
Doesn't support JSON targets


thejpster:
If you supply a target as a JSON file, the build.rs can't detect the target. This is because `target` is a full path - in my case starts this starts `C:\`, which != `thumbv`.

I think the fix is to take the `file_stem()` of the path and use that. This should work for both JSON targets and built-in targets (that don't contain `.`). If the built-in target does use `.` (e.g. `thumbv8-m.base`) we'll need to be a bit cleverer.

japaric:
> I think the fix is to take the file_stem() of the path and use that.

This doesn't always work. There's no guarantee the file name matches the LLVM name or it's even a proper triple, e.g. `foo.json` is a valid name for a custom specification file. The proper way is to run `rustc --print cfg --target $TARGET` and then parse the output. However, on beta / stable the output doesn't contain enough information to discern ARM Linux from ARM Cortex-M. For example, `target_feature="mclass"` only appears in the nightly output of `rustc --print cfg --target thumbv7m-none-eabi` and it's not stable, i.e. it may disappear on any nightly update.

If we want to "support" custom targets the documentation should state that this is best effort support and than nightly updates may break things.

thejpster:
I agree it requires the user to name their JSON file correctly, but at least they have that option. Currently, `cortex-m` and `cortex-m-rt` don't work at all with a JSON target file (i.e. thumbv8m)

PR #304, opened 2021-01-08
Allow using the crate with custom target JSON specs

jonas-schievink:
This should fix https://github.com/rust-embedded/cortex-m-rt/issues/145 (cc @thejpster)

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,10 +1,21 @@
-use std::env;
 use std::fs::{self, File};
 use std::io::Write;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
+use std::{env, ffi::OsStr};
 
 fn main() {
-    let target = env::var("TARGET").unwrap();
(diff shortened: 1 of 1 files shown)


adamgreig:
Idea seems good but this breaks when target is `thumbv8m.base-none-eabi` and similar.

jonas-schievink:
`Path::new("thumbv8m.base-none-eabi.json").file_stem().unwrap().to_str().unwrap()` results in `thumbv8m.base-none-eabi`, so this should work fine. `file_stem` only strips preceding directories and the last suffix.

adamgreig:
Won't it also run for just "--target thumbv8m.base-none-eabi", without a JSON spec, though?

jonas-schievink:
Oh, right, that's a problem.

adamgreig (approved):
Thanks!

bors merge

adamgreig (approved):
bors merge

PR #306, opened 2021-01-25
fix #74

spookyvision:
special-cased linker section names to satisfy macos linker constraints

Issue #74, opened 2018-05-16
LLVM error on High Sierra


chocol4te:
Building `cortex-m-rt` on `nightly`, `beta` and `stable` in Mac OS High Sierra (tested on two machines) results in the following error:

```
LLVM ERROR: Global variable '__RESET_VECTOR' has an invalid section specifier '.vector_table.reset_vector': mach-o section specifier requires a segment whose length is between 1 and 16 characters.
```

It build correctly on my Ubuntu 18.04 machine.

japaric:
> mach-o section specifier requires a segment whose length is between 1 and 16 characters

Are you compiling the `cortex-m-rt` crate for the host (for mac)?

If that's the case we can drop the link_section when compiling for mac; trying to link a cortex-m program compiled for the host (mac / Linux) is going to fail anyway.

chocol4te:
I'm so sorry, I'm the biggest idiot. I forgot to set the default target in .cargo/config.

jonas-schievink:
The cortex-m crate goes out of its way to build at least on Linux, would be nice if all the core crates consistently built on Linux, macOS, and Windows

japaric:
we ran into a similar issue in the `defmt` project.
we found out that on macOS the longest linker section name can be `.section,0123456789ABCDEF` (note: **comma** after `.section` not period and a max of 16 characters after the comma)
this issue can probably be fixed by adding a `cfg_attr(is_macos, link_section = "..")` to the relevant static variables

therealprof:
Indeed, I've encountered it frequently.

berkus:
What does this fix do? It replaces section name with exactly the same but taking half a kilobyte more space in source code; am I missing something there?

therealprof:
@berkus Not sure what you mean. What are you seeing and on which system? This is changing the section names for macOS only because it has a weird length restriction. It should not affect anything except to prevent linker errors on macOS.

berkus:
@therealprof ok,  lets go with the way of example, shall we.

```
#[cfg_attr(target_os = "macos", link_section = ".vector_table,interrupts")]
#[cfg_attr(not(target_os = "macos"), link_section = ".vector_table.interrupts")]
```

On  macOS the  section  name would be `.vector_table.interrupts`,  however on any other system it would be `.vector_table.interrupts` which is EXACTLY THE SAME.

So now you've reduced readability, understandability and maintainability of the  code to provide exactly the same section names for every system out there.

[...]

PR #310, opened 2021-01-27
Only emit link_section for cortex-m

adamgreig:
Previously we always emitted `link_section`, even though it only had an
effect when our linker script was being used (and only made sense on
cortex-m targets). This breaks building the code for a MacOS target,
which is occasionally useful for running `cargo check` etc.

In the macros crate we don't have the target information available, so
instead we continue to emit `link_section` except specifically on MacOS.

This keeps the fix from #306 but hopefully resolves the confusion in https://github.com/rust-embedded/cortex-m-rt/issues/74#issuecomment-768079657.

### DIFF
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -278,8 +278,10 @@ pub fn exception(args: TokenStream, input: TokenStream) -> TokenStream {
                 #(#attrs)*
                 #[doc(hidden)]
                 #[export_name = "HardFault"]
-                #[cfg_attr(target_os = "macos", link_section = ".HardFault,user")]
-                #[cfg_attr(not(target_os = "macos"), link_section = ".HardFault.user")]
+                // Only emit link_section when building for embedded targets,
+                // because some hosted platforms (used to check the build)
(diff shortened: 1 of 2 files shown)


therealprof:
Just curious, why do we need these sections at all then on a real OS? Can't we just generally omit them if we have a `target_os` defined?

adamgreig:
We don't need to emit any link sections on non-cortex-m platforms, which is what this PR does except for the macro crate. Do you have a better cfg match in mind for the macro crate? I'd just left it as `not(target_os="macos")` but if you have a better idea I'm all for it.

therealprof:
Ah, I was just looking at the `macro` part of the diff.

Would `target_os=""` work?

jonas-schievink:
`cfg(target_os = "none")` should work

adamgreig:
Yep, that seems to be working in my quick test (still places HardFault immediately after HardFaultTrampoline, while using target_os=dummy places it beforehand in with the other functions). Updated.

jonas-schievink (approved):
bors r+

PR #313, opened 2021-02-23
[wip] implement the new attribute syntax

jonas-schievink:


Dirbaio:
Should resources on interrupt handlers require Send? You can change the priority of an interrupt at runtime, effectively "sending" the resource values from one priority level to another.

Also, can the DefaultHandler have resources? I think that's unsound because it can run reentrantly if there are 2 interrupts at different priority without a handler set.

Issue #314, opened 2021-02-25
SCB.ICSR is more than 8 bits


Dirbaio:
This code here reads ICSR and casts it to 8 bits. In cortex-m3 it is more than 8 bits because there can be up to 480 interrupts. 

https://github.com/rust-embedded/cortex-m-rt/blob/ca4790f40738bcc770285c8080955a2077682078/macros/src/lib.rs#L230

This is the same as https://github.com/rust-embedded/cortex-m/issues/332

adamgreig:
Closed by #315.

PR #315, opened 2021-02-25
Use volatile read for ICSR register

lulf:
This prevents the compiler from optimizing the read.

Edit: I also added a change to include the 9th bit in the IRQ. I can leave that out if it would break anything. 

#314


adamgreig (approved):
Thanks, looks good! Keeping the top bit makes sense here since it was already an i16 variable type.

This will also close #314.

bors merge

adamgreig:
bors retry

PR #324, opened 2021-04-29
bump version to 0.6.13

jordens:
The currently released version is 0.6.13 but the version in Cargo.toml was still 0.6.11.
This enables cargo-patching in the git repository if other crates need newer cortex-m-rt.


adamgreig (approved):
Yea, worth doing. It feels like there should be a better way to have the git repo indicate it's ahead of the released 0.6.13 but here we are. At least the changelog in master reflects 0.6.13 and I believe all those 0.6.x changes are already included in master too.

bors merge

PR #330, opened 2021-07-12
Prepare for v0.7.0

adamgreig:
This is mostly a changelog update, including the updates from 0.6 and adding a few missing PRs too.

I haven't yet reviewed our open issues but I'm sure there are some low-hanging fruit we should get in for this release, does anyone have any suggestions? I'm happy to update this PR a few times if we can get a few more changes landed first.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,27 +11,35 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Add support for CMSE secure gateway veneers ([#297]).
 - Allow using the crate with custom target JSON specs ([#304]).
+- Export Exception enum for other crates to use ([#224])
 
+[#224]: https://github.com/rust-embedded/cortex-m-rt/pull/224
 [#297]: https://github.com/rust-embedded/cortex-m-rt/pull/297
 [#304]: https://github.com/rust-embedded/cortex-m-rt/pull/304
 
 ### Fixes
 
-- Various fixes to the linker script ([#265], [#286]).
+- Various fixes to the linker script ([#265], [#286], [#287], [#323]).
 - Use the correct ABI for the `main` symbol ([#278]).
 - Add barriers after FPU enabling ([#279]).
-- (ARMv6-M) Set LR value to a known value on reset (as the ARM spec requires)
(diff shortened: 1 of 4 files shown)


adamgreig:
Oh, great, one of our dev-dependencies, `compiletest_rs`, depends on `regex`, which depends on `memchr`, which has started using `non_exhaustive`, which breaks our MSRV. Is this a semver breaking change? Hard to say. It's not in our actual dependency chain, but I don't know that we can run the CI for the MSRV without it. We're currently 1.31; this would require a bump to 1.40.

Sympatron:
Since `compiletest_rs` depends on `regex ^1.0` since version `0.3.14` it automatically updated `regex <1.4.6` to `regex >=1.5.0` when it was released 2 month ago this can't really be changed. And it already is a breaking change since `regex 1.5.0` was released and updated from `memchr ^2.2.1` to `memchr ^2.4.0`.

The only solution would be to change `compiletest_rs` dependency on `regex` to `^1.4.0` to keep it's MSRV lower. This would be better anyway IMO.

eldruin:
I would have said the MSRV relates only to "cargo build", so not including `dev-dependencies`.

adamgreig:
I misread/misremembered the MSRV yesterday: we're actually on 1.39.0 now, so bumping to 1.40 for non_exhaustive wouldn't be a huge change for a breaking release. In a way I think @eldruin is right, really I'd say just being able to build it should be enough, but it's very hard to know if we actually still support the MSRV if we can't run the tests on it.

Possibly we could pin the compiletest_rs dependency to an earlier version (perhaps only when testing the MSRV)?

adamgreig:
Updated to skip the compiletest_rs tests on 1.39.0, but still run the build for all targets, which seems to have fixed it.

Happy to consider bumping MSRV to 1.40 instead if anyone has strong feelings.

adamgreig:
It's been a month and no suggestions for any more changes to bring in, so shall we call this 0.7? @rust-embedded/cortex-m ?

thejpster:
Fine by me

therealprof:
Ship it.

thejpster (approved):

adamgreig:
bors r=thejpster

PR #337, opened 2021-11-22
Fix unwinding through `Reset`

jonas-schievink:
Unwinders may detect the end of the program by seeing `0xFFFFFFFF` in `lr`, which is why code to ensure that it has that value was added in https://github.com/rust-embedded/cortex-m-rt/pull/293. However, the `bl main` overwrites that value with the current program counter. This PR saves the old `lr` value on the stack, and adds debuginfo entries to allow an external unwinder to restore the value.

This fixes https://github.com/knurling-rs/probe-run/issues/277


asm.S
```diff
@@ -94,6 +94,10 @@ Reset:
 #endif
 
 4:
+.cfi_def_cfa sp, 0
```
jonathanpallant, line 97: This might be worth an explanation, as it's not something you often come across in ARM asm.

Maybe link to https://sourceware.org/binutils/docs/as/CFI-directives.html, or include their note:

> `.cfi_def_cfa` defines a rule for computing CFA as: take address from register and add offset to it.

jonathanpallant (changes requested):

jonathanpallant:
Can confirm it resolves the upstream issue though:

```
PS C:\Users\Test\Documents\knurling\app-template> cargo run --bin hello
    Blocking waiting for file lock on build directory
   Compiling cortex-m-rt v0.7.0 (https://github.com/knurling-rs/cortex-m-rt.git?branch=fix-unwind-through-reset#9c88eab9)
   Compiling stm32f4xx-hal v0.10.1
   Compiling my_project v0.1.0 (C:\Users\Test\Documents\knurling\app-template)
    Finished dev [optimized + debuginfo] target(s) in 15.88s
     Running `probe-run --chip STM32F429ZITx target\thumbv7em-none-eabi\debug\hello`
(HOST) INFO  flashing program (7.93 KiB)
(HOST) INFO  success!
────────────────────────────────────────────────────────────────────────────────
Hello, world!
└─ hello::__cortex_m_rt_main @ src\bin\hello.rs:8
────────────────────────────────────────────────────────────────────────────────
(HOST) INFO  device halted without error
```

asm.S
jonas-schievink: Added a comment and link

jonathanpallant (approved):

jonathanpallant:
Ugh. Wearing the wrong hat.

thejpster (approved):

thejpster:
^ have switched hats

Issue #283, opened 2020-07-31
VTOR and initial stack for debuggers


debuti:
In `crt0.s` you'll usually find statements like the next one at the very beginning of the code:
`  ldr SP, =_stack_start`

To set the initial stack pointer to the correct value. You'll find it along with code to set `VTOR`, and other tasks considered as protective programming. This allows to mimic the HW reset behavior to start off in the same environment when the software is launched from a debugger (ie. lauterbach)

Can this be implemented in `fn Reset() -> !`? 

One can argue that this can be implemented inside the preinit function, but in the stack pointer case it wont work as the stack has already been used by that moment and on function epilogue it will be incremented causing an incorrect mem access. Setting it using the debugger is also possible, but is not very clean to have this values (IVT location or stack start) in the debugger context.

adamgreig:
See also #55 where this was discussed previously. I think there have been similar discussions on the chat before too but not sure if they made it into issues. Do you have a specific debugger where this is a problem? I don't understand why using a debugger should require changing VTOR. It seems like a correct debugger should ensure SP is initialised if it's jumping code execution to the reset vector manually; otherwise it could use the AIRCR register to trigger a local reset.

Previously it's been mentioned in the context of bootloaders which might fail to change VTOR (and SP, I suppose) before jumping into the application, but really that seems like a defect in the bootloader. User applications can change VTOR in `pre_init` but you're right that it's too late to update SP there.

On Cortex-M0 there's no support for VTOR and instead remapping the vector table is done in some chip-specific fashion, for example by remapping SRAM to address 0 and writing a vector table to the start of the SRAM. We can't implement that in cortex-m-rt because it's entirely application and MCU specific, and requires corresponding updates to the user's `memory.x` etc. It could be supported by a BSP or HAL perhaps.

In the context of bootloaders I was fairly sure we shouldn't try and fix them in cortex-m-rt, since it's very platform specific and we can't make all the right guesses, but I hadn't considered debuggers too. I'd be interested in hearing about why a debugger needs the user code to set up SP/VTOR for sure.

debuti:
Makes sense, but i think that at least the stack pointer -afaik common to all cortex m processors- can be set at reset, leaving the rest for the preinit, where it can be easily adjusted with some unsafe blocks.

[...]

Issue #254, opened 2020-03-20
Add force-setting of stack pointer in startup code


korken89:
During experimenting with new unsafe code I realized that the stack pointer can easily be corrupted, and this can propagate and stick after a system reset while having a debugger connected.
This causes `preinit` to automatically hardfault if using the stack and the zeroing/init of RAM under no optimization also directly hardfaults.

I propose we add the code to always hard-load the stack pointer from the vector table.
It will cost a few bytes, but I argue that it is a price worth to pay to have stable debug/implementation runs.
Currently it is trivial to get into a very hard to debug state.

jonas-schievink:
How/Why does this situation happen, exactly? I wouldn't want to work around issues we don't fully understand in cmrt.

adamgreig:
The architecture reference manuals seem pretty explicit that SP is always initialised from the vector table on any power-on or reset, with no mention of exceptions, so something definitely seems wrong here. Do you have a short reproducible example?

korken89:
Sadly I'm far gone from this issue right now, but I'll look into reproducing it.
The issue is not on the Cortex-M core or the reset itself though, it's the debuggers. It seems some do not implement reset correctly, for example the stlinkv2, which after reset `mon reset halt` had the old stack pointer in some cases.

jonas-schievink:
I believe we discussed this at some point (or a similar issue) and concluded that this is a bug in the debugger that we don't really want to work around at this point, so closing this.

Issue #102, opened 2018-09-03
DefaultPreInit gets random address in debug mode [Stack pointer initialization issue]


korken89:
Hi all,

I have stumbled on a strange issue that took some work to find as it does not happen often.
My issue was that when compiling in debug mode the address provided by `__pre_init` becomes different addresses, depending on when the system was reset. And this comes from the stack pointer not being initialized yet, and a corrupt stack pointer passing through.
This can occur if the system's stack pointer has been corrupted and a soft-reset has been requested, then the erroneous value of the stack pointer sticks until after reset.

I wrongly assumed that the system always read the stack pointer from the vector table and wrote it, hence I recommend that we add an assembler instruction to set MSP and not allow this kind of undefined behavior. 
Comments @rust-embedded/cortex-m ?

Example of corrupt stack pointer after reset:
```
r0             0x8000251        134218321
r1             0xf00000 15728640
r2             0x20000000       536870912
r3             0x0      0
r4             0x0      0
r5             0x0      0
r6             0x0      0
r7             0x0      0
r8             0x0      0
r9             0x0      0
r10            0x0      0
r11            0x0      0
r12            0x0      0
sp             0xfffffe90       0xfffffe90
lr             0x8000315        0x8000315 <core::mem::uninitialized+12>
pc             0x80001c4        0x80001c4 <Reset+12>
xpsr           0x61000003       1627389955
fpscr          0x0      0
msp            0xfffffe90       0xfffffe90
psp            0x0      0x0
primask        0x0      0
basepri        0x0      0
faultmask      0x0      0
control        0x0      0
```

korken89:
Hmm, I think you are correct. I have seen it both when flashing an empty chip, or when restarting (load + continue) via the debugger after hardfault. Loading with a forced reset does indeed help, where I was assuming it was doing resets with load.

Still being able to arrive in this state at all is, for me, unacceptable. :)

adamgreig:
Sounds like an error in the debugger if it's jumping to the reset vector without either doing an actual reset or loading SP itself. Who knows what else is not correctly initialised? I usually `run` rather than `continue` after reflashing or hitting exception handlers which I _think_ does cause a reset.

I'm not super keen on having every single cortex-m-rt program begin by writing SP when that's exactly what the hardware should do anyway. What else would we have to do to ensure correct state? Any of the peripheral registers (both core and device-specific) could also be wrong...

ithinuel:
[...]

Issue #55, opened 2018-02-05
Support vector table relocation?


ah-:
Hi Jorge and friends,

For various reasons I'm trying to use cortex-m-rt on a stm32l151 with a bootloader at 0x0800_0000.
This means my memory.x looks like this:
```
MEMORY
{
  FLASH : ORIGIN = 0x08004000, LENGTH = 48K
  RAM : ORIGIN = 0x20000000, LENGTH = 16K
}
```
so my user code plus the vector table starts at 0x0800_4000.
On startup the bootloader runs first, and once it's done it jumps to the reset handler at *0x0800_4004.
So far this works, and my main (well actually RTFMs init and idle) runs as expected.
However, none of the interrupts work, as the stm32 still uses the bootloader vector table at 0x0800_0000.

To fix that I'm currently running `unsafe { p.core.SCB.vtor.write(0x4000) };` in `init`, after which everything runs just fine again.

Would it make sense to handle this vector table relocation in cortex-m-rt or should we consider this a bug of the bootloader?

If we want to fix it in cortex-m-rt, how could the fix look like? We'd need to write the relative offset from start of flash to where the vector table ends up, but as far as I can tell we don't know anymore that the original start of FLASH was at 0x8000_0000.

adamgreig:
> should we consider this a bug of the bootloader?

I would say yes; bootloaders should either set VTOR or copy the application vector table to RAM and remap it appropriately (e.g. on thumbv6 without VTOR), before jumping to application code. See for example [this article](http://www.keil.com/support/docs/3913.htm) from ARM/Keil. The bootloader has all the information it needs to do this properly, while cortex-m-rt doesn't in general.

jonas-schievink:
Can this be closed, or is there a use case where cortex-m-rt needs to be involved?

adamgreig:
I vote for closing; this is definitely behaviour the bootloader itself should take care of, and if it doesn't, chip/board support code can also take care of it. Doesn't need any special help from cortex-m-rt AFAIK.

jonas-schievink:
Okay, closing then. If some use case comes up we can always reopen.

PR #338, opened 2021-11-23
Set SP and VTOR in Reset vector

adamgreig:
I'm opening this PR for discussion after today's meeting, I'm not really sure I want it to be merged but I thought it might be helpful to see what it could look like.

## VTOR

The VTOR register offsets the vector table from the default 0x0000_0000 to a user-provided address, which is required if the user's vector table is not at the start of memory (for example, because a bootloader is present).

We occasionally get people asking for help because none of their interrupts are working and it turns out their bootloader does not set VTOR, so interrupts were not taking the right vector. Typically this is on platforms with manufacturer or vendor bootloaders rather than ones the user has installed, so they may not even be aware the bootloader is running.

It's a hard problem to debug and we could resolve it by having cortex-m-rt always write VTOR at startup, since it knows what the vector table address is. On ARMv6-M the VTOR register is optional, but if not present it must be RAZ/WI, which means it's safe to write to it even though it won't have any effect. On ARMv7-M and ARMv8-M VTOR is always implemented. In practice no Cortex-M0 core has VTOR, but many Cortex-M0+ cores do.

However, this write is useless for the vast majority of applications, using up flash space and boot time without any benefit. I think bootloaders should always be responsible for setting VTOR, and for example this [document from Arm](https://developer.arm.com/documentation/ka002218/latest) agrees, but clearly not all bootloader authors read it.

## SP

This seems even rarer, but we have also had reports of debuggers resetting the chip by jumping to the reset vector without setting SP (the stack pointer, which is normally set to the first word of the vector table by the CPU at boot). I don't know what else those debuggers aren't setting, but from reports it sounds like setting SP is enough to get some basic functionality. This seems even more cursed (what else should we set, if we can't rely even on this?) but again is pretty easy to include if we wanted to. It's possible some bootloaders also fail to set SP before jumping to user code, although I don't recall hearing about that and it seems even more egregious. It would probably be helpful to have some more specific references to situations where this is necessary (see related issues).

## Making this optional

At the moment the reset vector is pre-built and distributed as a binary, so there's no simple way to control this behaviour with feature gates (except for shipping a _lot_ of different binaries, one for each architecture for each set of features). In the future when `global_asm!` becomes stable, it will probably be easy to feature-gate this functionality, and then it could be pretty easily enabled by users who need it and not others. Since many of the common use cases are for specific platforms which typically have a commonly-used HAL, those HALs could enable the feature for their users, so many users would not need to know about VTOR.

So, one (good?) option is to wait for `global_asm!` and then revisit this.

## Before or after pre_init?

I've put the new instructions before `pre_init` so that user code calling `pre_init` could choose to override them. I don't think it's likely any user code would need to run before these instructions: there's only 5 of them, so it's unlikely e.g. a watchdog would fire, and they don't access RAM, so it's unlikely RAM would need enabling, and they only set up the chip to a state it should normally be in after boot (well, VTOR is usually 0, but what we write should be equivalent).

## Related issues

These topics have been requested/discussed before:

* #283 requesting pretty much exactly this addition, "considered as protective programming", "mimic the HW reset behavior to start off in the same environment when the software is launched from a debugger"
* #55 has a bootloader that doesn't set VTOR
* #254 STlink debugger not setting SP correctly
* #102 SP not being set or seen by a debugger correctly

And besides those issues, it comes up occasionally on the Matrix chat.


Dirbaio:
One drawback is this will break firmwares using Nordic's nRF softdevices. 

The softdevice lives at the start of flash, the application firmware is after it.. The softdevice handles some irqs, then forwards the rest to the application. Setting VTOR will break it, because all irqs go straight to the application so the softdevice won't be able to handle theirs.

I'd be OK with this being optional, opt-in though!

PR #341, opened 2022-01-19
Make the address of the vector table configurable.

david-boles:
Not sure if this is the right way to do it; please let me know if there's a better alternative. I'm trying to add dummy sections for a bootloader header and trailer, this change allows my `memory.x` file to be pretty clear:

```
MEMORY
{
  /* NOTE K = KiBi = 1024 bytes */
  FLASH : ORIGIN = 0x00008000, LENGTH = 464K
  RAM : ORIGIN = 0x20000000, LENGTH = 64K
}

SECTIONS
{
  .mcuboot_header 0x8000 :
  {
    FILL(0xAAAAAAAA)
    . = . + 32;
  } > FLASH
  PROVIDE(_svector_table = 0x8020);
}
INSERT BEFORE .vector_table;

SECTIONS
{
  .mcuboot_trailer . :
  {
    FILL(0xFFFFFFFF)
    . = . + 40;
  } > FLASH
}
INSERT AFTER .gnu.sgstubs;
```

### DIFF
--- a/link.x.in
+++ b/link.x.in
@@ -63,8 +63,11 @@ SECTIONS
   PROVIDE(_stack_start = ORIGIN(RAM) + LENGTH(RAM));
 
   /* ## Sections in FLASH */
+  
+  PROVIDE(_svector_table = ORIGIN(FLASH));
+  
   /* ### Vector table */
-  .vector_table ORIGIN(FLASH) :
+  .vector_table _svector_table :
   {
     /* Initial Stack Pointer (SP) value */
     LONG(_stack_start);


adamgreig:
This PR might be a good idea regardless, but as another option to solve your problem I wonder if you could do this? I've done something very similar in the past to leave room for configuration data before the vector table in bootloader applications.

```
MEMORY {
    HEADER : ORIGIN = 0x00008000, LENGTH = 32
    FLASH : ORIGIN = ORIGIN(HEADER) + LENGTH(HEADER), LENGTH = 464K - LENGTH(HEADER)
    RAM : ORIGIN = 0x20000000, LENGTH = 64K
}

SECTIONS {
    .mcuboot_header . : {
      FILL(0xAAAAAAAA)
      . = . + 32;
    } > HEADER
}

SECTIONS
{
  .mcuboot_trailer . : {
    FILL(0xFFFFFFFF)
    . = . + 40;
  } > FLASH
} INSERT AFTER .gnu.sgstubs;
```

david-boles:
adamgreig, that definitely works too and I'll probably go with that unless and until this gets merged. It would feel nicer if the header and trailer could be treated the same way though.

PR #343, opened 2022-01-23
Add note about repository move to README

adamgreig:
After this is merged I'll mark this repository as archived, and transfer remaining issues.

I think we'll have to close open PRs as they can't be transferred as-is.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,11 @@
 
 > Startup code and minimal runtime for Cortex-M microcontrollers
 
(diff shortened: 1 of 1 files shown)


adamgreig:
I've disabled requiring `bors` to pass before allowing merging; CI is broken due to a new Rust version in a way that is fixed in the cortex-m repository. There isn't much point backporting the fix, but it should now be possible to manually click the merge button after approving.

thejpster (approved):

